PR #4, opened 2015-06-10
Remove convenience initializers

wesj:
Swift wouldn't let me override with these convenience initializers. It required that I override the private one. Then when I overrode it, it would get confused with the convenience ones. I have no idea if this is a Swift bug or yours, but it seems easier to just have one initializer with default arguments anyway. Code completion will suck, but its less code to manage.

### DIFF
--- a/Deferred/Deferred.swift
+++ b/Deferred/Deferred.swift
@@ -18,19 +18,9 @@ public final class Deferred<T> {
     private var protected: LockProtected<Protected>
     private let defaultQueue: dispatch_queue_t
 
-    private init(value: T?, queue: dispatch_queue_t) {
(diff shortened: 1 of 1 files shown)


zwaldowski:
Hey, we're sorry to be slow in responding! With all the massive changes in Swift 2.0, Deferred has seen some reworking to match. `Deferred` is no longer a class, but there's a new set of protocols that should ease this kind of work of almost sub typing. We're working on "a cancellable `Deferred`" ourselves. You can check that current state of this new work in [v2.0b3](https://github.com/bignerdranch/Deferred/releases/tag/v2.0b3) and the branch at #29. If you still want something the lines of this PR, please file an issue or reopen/rebase this PR. Thanks!

PR #5, opened 2015-06-10
Make non-final so Deferred can be extended

wesj:
We have some db operations that return Deferreds. I wanted to make them cancellable, so I really wanted to return a Deferred with a cancel method that could call sqlite3_interrupt() when called. Unfortunately, Deferred is final, so you can't override it. I'm guessing you really want that, but wanted to forward the changes in case you wanted them.

### DIFF
--- a/Deferred/Deferred.swift
+++ b/Deferred/Deferred.swift
@@ -11,7 +11,7 @@ import Foundation
 // TODO: Replace this with a class var
 private var DeferredDefaultQueue = dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0)
 
-public final class Deferred<T> {
+public class Deferred<T> {
     typealias UponBlock = (dispatch_queue_t, T -> ())
     private typealias Protected = (protectedValue: T?, uponBlocks: [UponBlock])
 


jgallagher:
I don't think Deferred should be subclassed. Since it's generic over any `T`, there isn't really a good way to implement a cancel method. We've used a cancelable Deferred via composition, though; something like this that uses a [Result](https://github.com/antitypical/Result) to represent failures:

```swift
enum MyError {
    Cancelled
    // other error cases
}

public final class CancelableDeferred<T> {
    private let deferred: Deferred<Result<T, MyError>>

    // ... omitting initializers and methods you want to expose on a CancelableDeferred

    func cancel() {
        deferred.fillIfUnfilled(.Cancelled)
    }
}
```

wesj:
Unfortunately that breaks a bunch of nice chaining functionality we've written and use unless we expose the inner deferred to outside callers. I played with making a Deferrable protocol here so that Deferred could be implemented by others (or a hidden composition could be reflected to the outside):
https://github.com/wesj/Deferred/commit/1a410420fa03be05966aeedb758e4229468ad50a

But protocols with generics are pretty painful to use in Swift :(

zwaldowski:
What chaining functionality did you add? We might be willing to add it to the class proper.

wesj:
I haven't written a ton of it. Just some wrappers and operators around Deferred<Result<()>> I thought, but I do see some more utility functions that have slipped in :)
https://github.com/mozilla/firefox-ios/blob/master/Utils/DeferredUtils.swift

jgallagher:
It looks like all of those work on `Deferred<Result<T>>`, right? Are you making a subclass specific to that specialization? I've been wondering if we should add some support for that kind of thing.

zwaldowski:
Same comment as on #4. @jgallagher was prescient in identifying the need for `Deferred<Result<T>>`. :wink: Similarly, if you still need this, please file an issue. Cheers!

PR #13, opened 2015-07-29
Lock-free implementation

zwaldowski:
* Uses modern GCD features to eliminate locking
   - Significantly reduce threading overhead
   - Minor behavioral change: `fill(_:)` no longer happens synchronously.
   - Requires iOS 8. Consensus among BNR folk: we should tag a 1.0 before merging (if merging is what we want to do) for use on iOS 7 with CocoaPods/including Deferred.swift directly.
* Renames and combines several methods in the spirit of Swift 2 and the stdlib.
* A new `wait(_:)` primitive underlies both `value` and `peek()`, also allowing for block-with-timeout use.
* `any(_:)` and `all(_:)` now take any collection, such as the result of a `lazy(_:).filter(_:)` or so on.


zwaldowski:
Rebased against master.

jeremy-w:
Why not remove LockProtected.swift and its tests file?

zwaldowski:
I did not feel the need to remove API that is currently in use;
 LockProtected has been used in every project that's used Deferred so far.

[image: Big Nerd Ranch logo] <http://www.bignerdranch.com/>

   - *Zachary Waldowski*
   - Software Developer
   - Big Nerd Ranch <http://bignerdranch.com/>
   - zachary@bignerdranch.com
   - (843) 333-6836

jeremy-w:
It looks like dead code from the perspective of all the code remaining in Deferred. If it's effectively getting equal footing, we should maybe add some docs on it to the README as well?

zwaldowski:
Blocked on Swift 2.0 (requiring very few changes, see 71286bcd4c) due to rdar://22437691 ([OpenRadar](http://www.openradar.me/22437691)).

zwaldowski:
Rebased to depend on #25. Please review that one first.

zwaldowski:
Updated to fix an API contract invalidation. No longer blocked for Swift 2; separate PR incoming.

zwaldowski:
Closing now that all changes will be for Swift 2 (#30, #33, and #29). The Swift 1.0 version will stay as-is. Split out the discussion about `LockProtected` into #34.

PR #7, opened 2015-06-17
Testing improvements

zwaldowski:
* Re-enables demi-benchmarks to improve test coverage
* Gets project under 100% test coverage
* Fix idiomatic mutability warnings in test target

PR #14, opened 2015-07-29
LockProtected improvements

zwaldowski:
Depends on #7 and #12.

* Replaces use of GCD queues as a locking mechanism with a GCD semaphore.
   - GCD semaphores briefly spin, and so is in the same ballpark as `CASSpinLock` when actual readers/writer behavior isn't a concern.
* Without the capturing semantics of dispatch queue blocks, the locking protocol can be made to use `@noescape`
* Adds a `pthread_rwlock_t` implementation
   - An early version of this branch also had an `NSLock` conformance. SO. SLOW.

TODO:
- [ ] Consider if readers-writer behavior is actually necessary
- [ ] Optimization for setter critical path?
- [ ] Ponies?

### DIFF
--- a/Deferred/ReadWriteLock.swift
+++ b/Deferred/ReadWriteLock.swift
@@ -9,8 +9,8 @@
 import Foundation
 
 public protocol ReadWriteLock {
-    func withReadLock<T>(block: () -> T) -> T
-    func withWriteLock<T>(block: () -> T) -> T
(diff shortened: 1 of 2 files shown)


jgallagher:
@zwaldowski This looks good to me. Pushed up a couple small commits - want to glance over them before we merge?

zwaldowski:
Looks good to me.

PR #16, opened 2015-08-07
rewrites README

jeremy-w:
Original README is preserved as docs.

This README is less API documentation (which we should probably shift into doc comments?) and more an explanation for why Deferred exists, what problems it solves, and how to use it to cleanly handle cancellation of an operation that would fill the deferred.


zwaldowski:
It's good, reads well; does it target the current syntax or the #13 one?

jeremy-w:
I believe I wrote it against the current code rather than the future code, so we could land it prior to the revision.

README.md
```diff
@@ -1,149 +1,181 @@
...
-`Deferred` knows nothing about).
-
-An instance of `Deferred` can only be filled once. It is a programmer error to
-fill an already-filled `Deferred`, and this will result in a runtime trap. (The
-method `fillIfUnfilled` is available for conditional filling.)
-
-## Usage - Producer
+## Vital Statistics
+- What: Lets you work with values that haven't been determined yet,
+        like an array that's coming (one day!) from a web service call.
```
randomstep: `one day!` -> `eventually`.  It is funny but only if you already know what is meant. Better to be clearer here.

README.md
```diff
@@ -1,149 +1,181 @@
...
+- CocoaPods: Not yet, but feel free to submit a PR!
+
+## Intuition
+A `Deferred<Value>` is a value that might be unknown now
+but is expected to resolve to a definite `Value` at some time in the future.
+It is resolved by being "filled" with a `Value`.
+
+A `Deferred<Value>` represents a `Value`.
+If you just want to work with the eventual `Value`, use `upon`.
+If you want to produce a `Deferred<OtherValue>` from the `Deferred<Value>`,
```
randomstep: I only understood this sentence because I already have some understanding of Deferred, map, and bind. The previous sentence was a little clearer to guess at even if the reader weren't familiar with `upon` - use `upon` to have your code take action whenever the deferred gets its value. 

This one is more like `continue to chain other async work when the value comes back`. Which - is not a useful description, unlike yours. I almost didn't even comment here as I'm not sure if there is a better way to explain it concisely here, before the excellent examples below.

randomstep:
It seems to me that fillIfUnfilled is safer and thus ought to be the default. If necessary leave a way for people to shoot themselves in the foot, but don't load the metaphorical crashing gun. (This is more for @jgallagher than this PR.) 

Very nice write-up. :+1:

README.md
jeremy-w: Would it be better to end with the Intuition section? Or nod to the
confusion and give the reader guidance: "This section is a concise summary
of how this works. If you're confused, read the examples in the next
section, then come back."

El sábado, 15 de agosto de 2015, Step Christopher <notifications@github.com>
escribió:

> In README.md
> <https://github.com/bignerdranch/Deferred/pull/16#discussion_r37141022>:
>
> > +- What: Lets you work with values that haven't been determined yet,
> > +        like an array that's coming (one day!) from a web service call.
> > +- Who: John Gallagher <jgallagher@bignerdranch.com <javascript:_e(%7B%7D,'cvml','jgallagher@bignerdranch.com');>> wrote this.
> > +- Swift: v1.2 for now.
> > +- License: MIT
> > +- Carthage: Yup.
> > +- CocoaPods: Not yet, but feel free to submit a PR!
> > +
> > +## Intuition
> > +A `Deferred<Value>` is a value that might be unknown now
> > +but is expected to resolve to a definite `Value` at some time in the future.
> > +It is resolved by being "filled" with a `Value`.
> > +
> > +A `Deferred<Value>` represents a `Value`.
> > +If you just want to work with the eventual `Value`, use `upon`.
> > +If you want to produce a `Deferred<OtherValue>` from the `Deferred<Value>`,
>
> I only understood this sentence because I already have some understanding
> of Deferred, map, and bind. The previous sentence was a little clearer to
> guess at even if the reader weren't familiar with upon - use upon to have
> your code take action whenever the deferred gets its value.
>
> This one is more like continue to chain other async work when the value
> comes back. Which - is not a useful description, unlike yours. I almost
> didn't even comment here as I'm not sure if there is a better way to
> explain it concisely here, before the excellent examples below.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/bignerdranch/Deferred/pull/16/files#r37141022>.
>


-- 
--
Jeremy W. Sherman <jeremy@bignerdranch.com>

README.md
jeremy-w: That sort of rewriting goes great with a small PR. :)

El sábado, 15 de agosto de 2015, Step Christopher <notifications@github.com>
escribió:

> In README.md
> <https://github.com/bignerdranch/Deferred/pull/16#discussion_r37140977>:
>
> > -[Running Closures Upon Fulfillment](#upon)), but also enable some higher level,
> > -powerful composition techniques.
> > -
> > -All properties and methods on an instance of `Deferred` can safely be called from
> > -multiple threads simultaneously; a lock is used internally for synchronization.
> > -Obviously this does not guarantee thread-safety of the contained result (which
> > -`Deferred` knows nothing about).
> > -
> > -An instance of `Deferred` can only be filled once. It is a programmer error to
> > -fill an already-filled `Deferred`, and this will result in a runtime trap. (The
> > -method `fillIfUnfilled` is available for conditional filling.)
> > -
> > -## Usage - Producer
> > +## Vital Statistics
> > +- What: Lets you work with values that haven't been determined yet,
> > +        like an array that's coming (one day!) from a web service call.
>
> one day! -> eventually. It is funny but only if you already know what is
> meant. Better to be clearer here.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/bignerdranch/Deferred/pull/16/files#r37140977>.
>


-- 
--
Jeremy W. Sherman <jeremy@bignerdranch.com>

README.md
randomstep: I like the latter choice - just point out there are examples below.

PR #22, opened 2015-08-24
Documentation sprint

zwaldowski:
Derived from #13. Provides documentation for all members, reaching 100% coverage under [Jazzy](https://github.com/realm/jazzy).

I produced an HTML docset in the repo directory with:

```shell
~/Downloads/jazzy/bin/jazzy -m Deferred -g https://github.com/bignerdranch/Deferred -a "Big Nerd Ranch" -u https://github.com/bignerdranch --module-version=1.2.0 -r http://bignerdranch.github.io/Deferred/docs/
```

Have at it. :fire:

### DIFF
--- a/Deferred/Deferred.swift
+++ b/Deferred/Deferred.swift
@@ -187,6 +187,11 @@ extension Deferred {
     Begins another asynchronous operation with the deferred value once it
     becomes determined.
     
+    `flatMap` is similar to `map`, but `transform` returns another `Deferred`
+    instead of an immediate value. Use `flatMap` when you want this deferred
+    value to feed into another asynchronous fetch. You might hear this referred
+    to as "chaining" or "binding".
+    
     :param: queue A dispatch queue for starting the new operation on.
     :param: transform A function to start a new deferred given the receiving
     value.
@@ -205,6 +210,9 @@ extension Deferred {
     
     /**
     Transforms the deferred value once it becomes determined.
+    
+    `map` executes a transform immediately when the deferred value is
+    determined.
     
     :param: queue A dispatch queue for executing the transform on.
     :param: transform A function to create something using the deferred value.
--- a/Deferred/Deferred.swift
+++ b/Deferred/Deferred.swift
@@ -191,8 +191,7 @@ extension Deferred {
     :param: transform A function to start a new deferred given the receiving
     value.
     
-    :returns: A new deferred value that is determined once the receiving
-    deferred is determined by beginning some new operation using that value.
+    :returns: The new deferred value returned by the `transform`.
     **/
     public func flatMap<NewValue>(upon queue: dispatch_queue_t = DeferredDefaultQueue, transform: Value -> Deferred<NewValue>) -> Deferred<NewValue> {
         let d = Deferred<NewValue>()
--- a/Deferred/Deferred.swift
+++ b/Deferred/Deferred.swift
@@ -32,8 +32,6 @@ public enum Timeout {
     /// Wait indefinitely.
     case Forever
     /// Wait for a given number of seconds.
-    ///
-    /// :note: An `.Interval(0.0)` will achieve different results from `.Now`.
     case Interval(NSTimeInterval)
 
     private var rawValue: dispatch_time_t {
--- a/Deferred/Deferred.swift
+++ b/Deferred/Deferred.swift
@@ -25,9 +25,15 @@ private var DeferredDefaultQueue: dispatch_queue_t {
     return dispatch_get_global_queue(QOS_CLASS_UTILITY, 0)
 }
 
+/// An amount of time to wait for a deferred value.
 public enum Timeout {
+    /// Do not wait at all.
     case Now
+    /// Wait indefinitely.
     case Forever
+    /// Wait for a given number of seconds.
+    ///
+    /// :note: An `.Interval(0.0)` will achieve different results from `.Now`.
     case Interval(NSTimeInterval)
 
     private var rawValue: dispatch_time_t {
@@ -42,6 +48,8 @@ public enum Timeout {
     }
 }
 
+/// A deferred is a value that may become determined (or "filled") at some point
+/// in the future. Once a deferred value is determined, it cannot change.
 public struct Deferred<Value> {
     private let accessQueue: dispatch_queue_t
     private let onFilled: dispatch_block_t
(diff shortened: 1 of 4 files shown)


jeremy-w:
The script in your PR message seems like a good candidate for adding to a Makefile.

Deferred/Deferred.swift
```diff
@@ -25,9 +25,15 @@ private var DeferredDefaultQueue: dispatch_queue_t {
...
 
+/// An amount of time to wait for a deferred value.
 public enum Timeout {
+    /// Do not wait at all.
     case Now
+    /// Wait indefinitely.
     case Forever
+    /// Wait for a given number of seconds.
+    ///
+    /// :note: An `.Interval(0.0)` will achieve different results from `.Now`.
```
jeremy-w: How do the results differ?

Deferred/Deferred.swift
```diff
@@ -62,21 +70,38 @@ public struct Deferred<Value> {
...
+    /// * In -O builds (the default for a "Release" configuration), program
+    ///   execution will stop.
+    /// * In -Ounchecked builds, the programming error is assumed to not exist.
+    ///
+    /// If your deferred requires multiple potential fillers to race, you may
+    /// disable the precondition.
+    ///
+    /// :param: value The resolved value of the deferred.
+    /// :param: assertIfFilled If `false`, race checking is disabled.
     public func fill(value: Value, assertIfFilled: Bool = true, file: StaticString = __FILE__, line: UWord = __LINE__) {
```
jeremy-w: The parameter `assertIfFilled` alone leaves me confused as to whether it's "assert if this `fill` call fills it" or "assert if it's already filled". It's clear from the comment and some thinking that it's the latter, but is there some alternative parameter name we could choose that would make it immediately clear at call sites, too?

Deferred/Deferred.swift
```diff
@@ -91,7 +116,16 @@ public struct Deferred<Value> {
             }
         }
     }
-
+    
+    /**
+    Call some function once the value is determined.
+    
+    If the value is already determined, the function will be submitted to the
+    queue immediately. An `upon` call is alreadys executed asnychronously.
```
jeremy-w: Not-a-word and typo: "alreadys executed asnychronously"

"Always"?

Deferred/Deferred.swift
```diff
@@ -116,16 +159,43 @@ public struct Deferred<Value> {
...
     }
-
+    
+    /**
+    Waits for the value to become determined, then returns it.
+    
+    This is equivalent to unwrapping the value of calling `wait(.Forever)`, but
+    may be more efficient.
+    
+    This getter will unnecessarily block execution. It might be useful for
```
jeremy-w: Is there a way we could apply a warning to uses of this? Would we even want to discourage its use so decisively?

Deferred/Deferred.swift
```diff
@@ -116,16 +159,43 @@ public struct Deferred<Value> {
...
     }
 }
 
 extension Deferred {
+    /**
+    Begins another asynchronous operation with the deferred value once it
+    becomes determined.
+    
+    :param: queue A dispatch queue for starting the new operation on.
+    :param: transform A function to start a new deferred given the recieving
```
jeremy-w: typo "recieving"

Deferred/Deferred.swift
```diff
@@ -116,16 +159,43 @@ public struct Deferred<Value> {
...
+    value.
+    
+    :returns: A new deferred value that is determined once the recieving
```
jeremy-w: typo "recieving"

More generally, I found this `:returns:` sentence really hard to follow. I think it might help to not talk about resolving the returned deferred. Maybe just, "The deferred value returned by `transform`."?

Deferred/Deferred.swift
```diff
@@ -135,7 +205,15 @@ extension Deferred {
...
         return d
     }
-
+    
+    /**
+    Transforms the deferred value once it becomes determined.
+    
+    :param: queue A dispatch queue for executing the transform on.
+    :param: transform A function to create something using the deferred value.
+    :returns: A new deferred value that is determined once the recieving
```
jeremy-w: "recieving" - I'd recommend just running a gsub across the whole project for "reciev" => "receiv" to catch all these

Deferred/Deferred.swift
```diff
@@ -135,7 +205,15 @@ extension Deferred {
         }
         return d
     }
-
+    
+    /**
+    Transforms the deferred value once it becomes determined.
```
jeremy-w: I think a `:note:` here and on `flatMap` to point out the distinction between them and why you'd use one vs the other might be very helpful. The param and returns docs are identical right now, which further confuses the issue.

Deferred/Deferred.swift
```diff
@@ -146,11 +224,26 @@ extension Deferred {
 }
 
 extension Deferred {
+    /**
+    Composes the recieving value with another.
```
jeremy-w: I just noticed that this isn't quite the same as `all([one, other])` because it returns a tuple rather than an array. :\

Deferred/DispatchBlockFlags.swift
```diff
@@ -10,24 +10,29 @@ import Dispatch
...
     public static var allZeros: dispatch_block_flags_t {
         return dispatch_block_flags_t(0)
     }
 
+    /// Convert from a bitmask of `UInt`.
     public init(rawValue: UInt) {
         self.init(rawValue)
     }
 
+    /// Create an instance initialized with `nil`.
```
jeremy-w: The docs on this differ in style from the docs in the rest of Deferred. These use an imperative ("Convert!", "Create!"), the others use a third-person indicative ("Converts", "Creates").

Deferred/LockProtected.swift
```diff
@@ -8,25 +8,45 @@
 
 import Foundation
 
+/// A `LockProtected` holds onto a value of type T, but only allows access to it
+/// from within a locking statement. This prevents accidental unsafe access when
+/// thread safety is desired..
```
jeremy-w: double `.` at the end here

Deferred/LockProtected.swift
```diff
@@ -8,25 +8,45 @@
...
     public func withReadLock<U>(block: T -> U) -> U {
         return lock.withReadLock { [unowned self] in
             return block(self.item)
         }
     }
 
+    /**
+    Give write access to the item within the given function.
+    
+    :param: block A function that writes to the contained item, and returns some
```
jeremy-w: Should we rename this `transform` in keeping with similar closure names in the Deferred API?

Deferred/DispatchBlockFlags.swift
zwaldowski: These are replicating the documentation from the stdlib. I'll change them, just pointing that out.

Deferred/Deferred.swift
zwaldowski: There might be some warnings in 2.0. All the `@availability`/`@available` that I know of cause errors.

Deferred/Deferred.swift
zwaldowski: It also, notably, allows heterogenous combination.

Deferred/DispatchBlockFlags.swift
jeremy-w: Maybe we should change the other docstrings to conform with the stdlib style, then?

Deferred/Deferred.swift
jeremy-w: Ah, that is nice. I'm used to Obj-C-derived types where you can always just throw them all in the `AnyObject` bucket, but that is an issue for actual Swift code.

Deferred/DispatchBlockFlags.swift
zwaldowski: I don't think it's necessary for all documentation to be in line with the Swift docs, we should just be consistent on our own and ideally fit some shade of BNR voice.

PR #25, opened 2015-08-26
Modernized front-facing API

zwaldowski:
* Renames and combines several methods in the spirit of Swift 2 and the stdlib.
* `any(_:)` and `all(_:)` now take any collection, such as the result of a `lazy(_:).filter(_:)` or so on.
   - `all` now short-circuits on empty collections.
* Increases documentation coverage
* For improved debugging, multiple `fill`s now trap in the caller's code, not in Deferred.
* Small performance improvement in `map`.

### DIFF
--- a/Deferred/Deferred.swift
+++ b/Deferred/Deferred.swift
@@ -12,31 +12,50 @@ private var DeferredDefaultQueue: dispatch_queue_t {
     return dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0)
 }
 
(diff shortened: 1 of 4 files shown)


zwaldowski:
I developed a workaround for rdar://22437691, so #13 should be fine, but for cleanliness' sake let's go one-at-a-time. This PR should be good-to-go based on the review from #13.

zwaldowski:
Updated to include documentation pass (#22).

zwaldowski:
Outmoded by Swift 2. The #30, #33, and #29 all make more significant API changes.

PR #26, opened 2015-08-28
Initial Podspec for v2.0b1

lyricsboy:
For review. Thinking we could use @zwaldowski's swift 2 branch as a 2.0b1 release.

### DIFF
--- a/Deferred.podspec
+++ b/Deferred.podspec
@@ -1,0 +1,104 @@
+#
+#  Be sure to run `pod spec lint Deferred.podspec' to ensure this is a
+#  valid spec and to remove all comments including this before submitting the spec.
+#
+#  To learn more about Podspec attributes see http://docs.cocoapods.org/specification.html
+#  To see working Podspecs in the CocoaPods repo see https://github.com/CocoaPods/Specs/
+#
+
+Pod::Spec.new do |s|
+
+  # ―――  Spec Metadata  ―――――――――――――――――――――――――――――――――――――――――――――――――――――――――― #
+  #
+  #  These will help people to find your library, and whilst it
+  #  can feel like a chore to fill in it's definitely to your advantage. The
+  #  summary should be tweet-length, and the description more in depth.
+  #
+
+  s.name         = "Deferred"
+  s.version      = "2.0b1"
+  s.summary      = "An implementation of OCaml's Deferred for Swift."
+
+  s.description  = <<-DESC
+  Deferred is an asynchronous promise-style API that can be used as an
+  alternative to the "block callback" pattern.
+                   DESC
+
+  s.homepage     = "https://github.com/bignerdranch/Deferred"
+
+
+  # ―――  Spec License  ――――――――――――――――――――――――――――――――――――――――――――――――――――――――――― #
+  #
+  #  Licensing your code is important. See http://choosealicense.com for more info.
+  #  CocoaPods will detect a license file if there is a named LICENSE*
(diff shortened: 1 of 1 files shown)


Deferred.podspec
```diff
@@ -0,0 +1,104 @@
...
+  #
+  #  Specify the authors of the library, with email addresses. Email addresses
+  #  of the authors are extracted from the SCM log. E.g. $ git log. CocoaPods also
+  #  accepts just a name if you'd rather not provide an email address.
+  #
+  #  Specify a social_media_url where others can refer to, for example a twitter
+  #  profile URL.
+  #
+
+  s.authors             = ["John Gallagher", "Zachary Waldowski"]
```
zwaldowski: We should have our BNR email addresses in here using the `{ "Name" => "Email", … }` syntax.

zwaldowski:
We don't need the template comments, or at least not the parts that we don't use.

Deferred.podspec
```diff
@@ -0,0 +1,104 @@
...
+  s.osx.deployment_target = "10.10"
+
+
+  # ――― Source Location ―――――――――――――――――――――――――――――――――――――――――――――――――――――――――― #
+  #
+  #  Specify the location from where the source should be retrieved.
+  #  Supports git, hg, bzr, svn and HTTP.
+  #
+
+  s.source       = { :git => "https://github.com/bignerdranch/Deferred.git", :tag => "v2.0b1" }
```
zwaldowski: They generally recommend you derive the tag from the version to prevent mistaken pod pushes. `:tag => "v#{s.version}"`.

Deferred.podspec
```diff
@@ -0,0 +1,104 @@
...
+  #   mac_spec.platforms = [:osx]
+  # end
+
+  # ――― Project Settings ――――――――――――――――――――――――――――――――――――――――――――――――――――――――― #
+  #
+  #  If your library depends on compiler flags you can set them in the xcconfig hash
+  #  where they will only apply to your library. If you depend on other Podspecs
+  #  you can include multiple dependencies to ensure it works.
+
+  # s.requires_arc = true
```
zwaldowski: I think we need this active.

Deferred.podspec
```diff
@@ -0,0 +1,104 @@
...
+  #
+  #  CocoaPods is smart about how it includes source code. For source files
+  #  giving a folder will include any swift, h, m, mm, c & cpp files.
+  #  For header files it will include any header in the folder.
+  #  Not including the public_header_files will make all headers public.
+  #
+
+  s.source_files  = "Deferred"
+
+  s.public_header_files = "Deferred/Deferred.h"
```
zwaldowski: We shouldn't need this, the umbrella header is empty.

Deferred.podspec
lyricsboy: Certainly can't hurt, but I assumed that the nature of Swift would require it. Perhaps it's best to leave it on in the event some ObjC gets added? :shrug_emoji:

Deferred.podspec
lyricsboy: Worth leaving it in just in case we add stuff to that header later, or no?

zwaldowski: Yeah, on second thought we'd need it for the GCD impl. because bridging headers blah blah.

Deferred.podspec
zwaldowski: I've seen it on every Seift project with a pod, but clearly it doesn't affect anything.

lyricsboy:
Check it out, addressed your comments.

Issue #10, opened 2015-07-13
Deferred doesn't participate in QoS Overrides


zwaldowski:
From WWDC 2015 Session 212:

![screen shot 2015-07-13 at 1 43 42 pm](https://cloud.githubusercontent.com/assets/170812/8656377/347a057c-2965-11e5-8540-7aaf54eea1b6.png)

Deferred currently uses `dispatch_group_wait` and `dispatch_async` for chaining, which can exclude participation in QoS propagation and contribute to priority inversion issues. Use of the `dispatch_block` APIs should be investigated for `Deferred` chaining.

zwaldowski:
Since `value` uses `dispatch_block_wait` in certain implementations, this is resolved for that in #13 and #30 for Swift 1 and 2, respectively.

Supporting it for `upon` would be a trivial amount of extra code. However, for Swift 2.0, that is currently blocked by rdar://22437691 ([OpenRadar link](http://www.openradar.me/radar?id=5055516279570432)), and can't be easily fixed in the same way as in #30.

zwaldowski:
Resolved for Swift 2.0 (incl. workarounds) in #30 and #33.

lyricsboy:
Can we close this now that #30 is merged?

zwaldowski:
Indeed!

PR #30, opened 2015-09-02
Lock-free implementation

zwaldowski:
#13, but for Swift 2.0.

This is a long running PR with lots of changes; they haven't been removed or rebased out to show the paths taken. Please prefer the [Files changed](https://github.com/bignerdranch/Deferred/pull/30/files) view over commit-by-commit.

Changes are as follows:
* Significantly reduces threading overhead:
   - Uses atomics and `ManagedBuffer` for lock-free `fill`.
   - Uses modern GCD features to eliminate locking for `upon`.
* A new `wait(_:)` primitive underlies both `value` and `peek()`, allowing for block-with-timeout use.
* `any(_:)` and `all(_:)` now take any collection, such as the result of a `lazy.filter(_:)` or so on.
   - These cannot be made protocol extensions until there is a protocol defining the needed elements of `Deferred`. This is planned for a future PR.
* Removes any dependency on Foundation
* Adopts quality of service classes (#10)
* Makes the `value` getter for internal testing only (#31)


zwaldowski:
Updated to not expose the `deferred_*` C APIs to the public.

zwaldowski:
Updated to better work around `dispatch_block_t` bugs in Swift 2.0. All the Objective-C introduced by this PR has been removed.

zwaldowski:
After some thought, I've updated to use `qos_class_self()` to retrieve a generic queue when not providing an `upon` queue. (This is also used for the no-argument `map` and `flatMap`, as well as for rejoining `all(_:)`.)

Consider a scenario where a user taps on a button that performs a web request. Parsing the resulting JSON blob (via `map`) and upserting the JSON into Core Data (via `flatMap`) should both execute at a User Initiated QoS, as the user might be waiting on the work to complete.

zwaldowski:
Updated to clean up branch history.

DeferredTests/DeferredTests.swift
```diff
@@ -305,4 +320,63 @@ class DeferredTests: XCTestCase {
...
+        XCTAssertTrue(d.isFilled)
+    }
+
+    // The QoS APIs do not behave as expected on the iOS Simulator, so we only
+    // run these tests on real devices. This check isn't the most future-proof;
+    // if there's ever another archiecture that runs the simulator, this will
+    // need to be modified.
+    //
+    // TODO: Add a clause for a tvOS test target
+    #if os(OSX) || (os(iOS) && !(arch(i386) || arch(x86_64)))
```
lyricsboy: I thought there must be a better way to handle this, and you probably tried, but I'm seeing mysterious behavior so I'll comment here anyway.

Things like `TARGET_OS_SIMULATOR` as defined in `TargetConditionals.h` are available to use in Swift, but from my testing:

1. using it with compile-time `#if` does not pass in the iPhone Simulator
1. using it with regular `if` at runtime *does* pass in the Simulator

What's up with that?

zwaldowski: Simple `#define`s from C are only available as `let` symbols in Swift; they are not available for compile-time checking. We might be able to define a convoluted `#if` flag using build configurations, though @jgallagher and I determined when this test was written for internal code that this was the best solution.

Deferred.xcodeproj/project.pbxproj
```diff
@@ -385,6 +391,7 @@
 			baseConfigurationReference = DB32954F1B680744002160CC /* Debug.xcconfig */;
 			buildSettings = {
 				ENABLE_TESTABILITY = YES;
+				ONLY_ACTIVE_ARCH = YES;
```
hitsvilleusa: I trust that this is changed from the default for a good reason.  Just calling it out in case it was accidental.

zwaldowski:
@hitsvilleusa: I'd recommend looking at the "Files changed" view for the whole PR instead. Everything you've commented on so far has changed or been removed.

Deferred.xcodeproj/project.pbxproj
zwaldowski: Xcode 7.1 insisted. It's in the xcconfig but it would moan every time you opened the project.

Deferred/Deferred.swift
```diff
@@ -6,39 +6,131 @@
 //  Copyright © 2014-2015 Big Nerd Ranch. Licensed under MIT.
 //
 
-// This cannot be a class var, new storage would be created for every
-// specialization. It also could not be used as a default argument as it is now.
-private var DeferredDefaultQueue: dispatch_queue_t {
-    return dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0)
+import Dispatch
+
+private final class Storage<T> {
```
hitsvilleusa: I would love some comment indicating the reason for this classes existence.

Deferred/LockProtected.swift
```diff
@@ -6,8 +6,6 @@
 //  Copyright © 2014-2015 Big Nerd Ranch. Licensed under MIT.
 //
 
-import Foundation
-
```
hitsvilleusa: I don't see this file being used by anything other than tests.  If it isn't needed, I think it should be removed from the repo.

Deferred/ReadWriteLock.swift
```diff
@@ -6,7 +6,7 @@
 //  Copyright © 2014-2015 Big Nerd Ranch. Licensed under MIT.
 //
 
-import Foundation
+import Dispatch
```
hitsvilleusa: I don't see this file being used by anything other than tests.  If it isn't needed, I think it should be removed from the repo.

Deferred/Deferred.swift
```diff
@@ -6,39 +6,131 @@
...
+//  - The Element buffer has a stable pointer when locked to a single element.
+//  - Better holdsUniqueReference support allows for future optimization.
+private final class DeferredBuffer<Value, SideData>: ManagedBuffer<SideData, Storage<Value>?> {
+    
+    static func create(sideData: () -> SideData) -> DeferredBuffer<Value, SideData> {
+        return create(1, initialValue: { _ in sideData() }) as! DeferredBuffer<Value, SideData>
+    }
+    
+    deinit {
+        // super's init automatically destroys the Value
```
hitsvilleusa: deinit?

Deferred/Deferred.swift
zwaldowski: As noted on the similar comment, will do.

Deferred/LockProtected.swift
zwaldowski: Discussed in the comments on #13 (the Swift 1.x version of this PR). We should discuss in Slack in the morning.

/cc @jeremy-w

Deferred/ReadWriteLock.swift
zwaldowski: Same note w.r.t. `LockProtected`.

Deferred/LockProtected.swift
zwaldowski: Also cc @jgallagher

Deferred/Deferred.swift
zwaldowski: Yes?

jeremy-w: I think the point was that the comment says "super's *init* automatically destroys the value" vs "super's *deinit"…".

zwaldowski: Ah. Thanks!

Deferred/Deferred.swift
```diff
@@ -6,39 +6,131 @@
...
     public init() {
-        self.init(nil)
+        storage.initializeWith(nil)
     }
     
     /// Initialize a filled Deferred with the given value.
     public init(value: Value) {
-        self.init(value)
+        storage.initializeWith(value)
+        markFilled(storage.value)
```
hitsvilleusa: I find this frustrating to read.

As a reader of the code, I think "Ok ... this Deferred type must manage a value, I expect to see that value as a private property or something similar" yet, even as I read something as simple as this `init` method, I think hmm ... "OK ... not quite what I was anticipating, but ok ... it looks like there is some `Storage` type that must manage the actual value ... not sure why ... but I can accept that.  Then, since I immediately provided a value it "marks as filled" with my value ... ok."

Yet, the last thought is in fact not true.  The call to `storage.value` is *not* the value that I passed in but a `dispatch_block_t`.  This is not what I would expect as I just read these two lines of code.

I am not sure what I would suggest here.  Comments might help?  Or, perhaps some indication that the `value` passed to `initializeWith(value)` is not the same as `storage.value`?

zwaldowski: The `.value` comes from the buffer type. I'll look at making the naming more clear; might just make an alias.

hitsvilleusa: I think what these two lines are making more clear to me about my resistance to these changes is that as I read the code, while I trust that it is correct, it is not remotely close to what I would have imagined it would look like.  This doesn't make it wrong, in fact, it might be very correct, it just makes it really tough to maintain as it is not obvious.

A possible suggestion would be more documentation as to the high level approach that you are taking, and why it is the best approach.

zwaldowski: I don't understand quite what you're asking. This topic has come up more than a few times; what do you imagine it would look like?

Deferred/Deferred.swift
```diff
@@ -6,39 +6,131 @@
...
     }
-    
+
     /// Check whether or not the receiver is filled.
     public var isFilled: Bool {
-        return protected.withReadLock { $0.protected != nil }
+        return dispatch_block_testcancel(storage.value) != 0
```
hitsvilleusa: This is another example of the same thing as my last comment about more documentation about the general approach.

I look at the line that was replaced, and I can make sense of it.

Here is my naive thought process as I take a look at this method for the first time, with the old line:

Before I even open the code I think:

1. I bet this method will just look at whether or not is has a value stored yet for the private property that it must manage, that property must be an optional.
2. I haven't thought about all of the scenarios, but I bet there are some threading concerns there, so I wouldn't be surprised to see something related to threading.
3. I open the code and look at the method.
4. `protected.withReadLock` ... ok, that makes some sense ... must be dealing with some threading concerns.
5. I don't know exactly what this `protected` thing is, but it appears to be dealing with locking/threads so that's enough insight for me.
5. Yep .. then it checks the optional value to see if it is nil or not ... ok, makes sense to me.

(The above is just what goes on in my mind as a potential maintainer of this code.)

When I look at the new code, I still have the same thoughts #1-3 as I am about to read the code, but then I look at the code and just feel like "what did I miss?".  "Maybe I am looking at the wrong thing?"

Again, I am not at all suggesting that the code is incorrect.  I am simply concerned that it is more difficult to understand the motivation.  The earlier suggestion of more documentation might help ease the maintenance of this class.

Deferred/Deferred.swift
hitsvilleusa: If I knew nothing about `Deferred` and had never opened the source code before, my initial, naive expectation is that there would be something like this:

```swift
private var value: T? = nil

public init(value: value) {
  self.value = value
}
```

randomstep:
:+1: from me. Once I understood the ManagedBuffer backing class, and the (confusing) use of value vs. sidecar data vs elements that flipflops through the methods a bit, I am able to follow the implementation. This could be as simple as a README or note above the class to understand `ManagedBuffer` before working on implementation of our buffer class, or it could involve renaming some of the "value" variables to be clearer. 

This could be filed as an Issue and done as a future PR, or done now before this is merged. I prefer whichever one gets this merged quicker :), as it has been hanging out long enough and can be improved quite reasonably by more PRs in the future.

Note: The branch needs to be updated from the base branch before it can be merged.

Deferred/Deferred.swift
jeremy-w: Bounce over to [where it's canceling](https://github.com/bignerdranch/Deferred/pull/30/files#diff-84698bee4c3154d49155aa8e9a06b3a2R112) with `dispatch_block_cancel`, and you can see the other half of the handshake. It looks like we're gradually pushing more of the grody threading concerns into GCD.

Deferred/Deferred.swift
jeremy-w: There is some high level documentation already, it's just in the PR (check the main discussion page with the PR description) rather than the project repo:

> - Significantly reduces threading overhead:
>  - Uses atomics and ManagedBuffer for lock-free fill.
>  - Uses modern GCD features to eliminate locking for upon.

Deferred/Deferred.swift
```diff
@@ -57,52 +149,28 @@ public struct Deferred<Value> {
...
-            if data.protected == nil {
-                data.uponBlocks.append( (queue, function) )
-            }
-            return data.protected
-        }
-        if let value = maybeValue {
-            dispatch_async(queue) { function(value) }
-        }
+    public func upon(queue: dispatch_queue_t = genericQueue, body: Value -> ()) {
+        _ = upon(queue, options: DISPATCH_BLOCK_INHERIT_QOS_CLASS, body: body)
```
hitsvilleusa: Just curious why the `_ =` is needed?

Deferred/Deferred.swift
```diff
@@ -110,15 +178,48 @@ public struct Deferred<Value> {
...
+
+    /**
+    Waits synchronously for the value to become determined.
+
+    If the value is already determined, the call returns immediately with the
+    value.
+
+    :param: time A length of time to wait for the value to be determined.
+    :returns: The determined value, if filled within the timeout, or `nil`.
+    */
```
hitsvilleusa: Just noting that the comment style changes from `///` to `/** */`.

jeremy-w: Thumbs up :+1: for the less noisy `/** … */` style. ;)

Deferred/Deferred.swift
jeremy-w: It might have a `WARN_UNUSED_VALUE` jobby attached, or it's just to let the reader know that the value is intentionally being discarded, similar to doing `(void)printf("foo\n");` to go "yes, this returns something, but I am intentionally ignoring that fact".

zwaldowski: It triggered an unused value warning, yeah.

zwaldowski:
Cleaned up naming and fixed mergability with base branch.

zwaldowski:
Also added some extra contents.

Deferred/Deferred.swift
zwaldowski: #33 completes the move to the new comment style. The `///` style is preferred by Swift 2. It triggers the new keyword syntax (` - returns:`) and has much clearer indentation style.

Deferred/Deferred.swift
zwaldowski: Duped some of that into the implementation

zwaldowski:
:construction_worker: approves. Give it another once-over and [file issues](https://github.com/bignerdranch/Deferred/issues/new) for the rest. Cheers!

Deferred/Deferred.swift
hitsvilleusa: Good to know.  Thanks!

hitsvilleusa:
:+1: I will merge first thing in the morning, assuming no other comments.

Nice work on this.

Issue #24, opened 2015-08-24
Ignoring semantics


zwaldowski:
The [original Deferred](https://ocaml.janestreet.com/ocaml-core/108.00.02/doc/async_core/Deferred.html) describes `unit` and `all_unit`, which produce `Deferred<Void>`. This has been useful in our own apps and for derived types.

A simple implementation could be no different from:

```swift
deferred.map { _ in }
```

However, there might be some optimization that could be made in the GCD implementation (#13) by reusing the queue and block or refusing to create storage. Some effort would ideally be made to prevent that derived `Deferred` from being filled at all.

zwaldowski:
Finished in #33.

Issue #21, opened 2015-08-19
provide a context-specific alias for flatMap to assist in learning the API


jeremy-w:
`flatMap` is a generic term that's great for people who can come at this from that level already, but does a disservice to people whose first exposure to something like this is `Deferred`. Providing a more deferred-specific API with a more suggestive name alongside `flatMap` would hopefully allow them to ramp up to understanding `flatMap` in this context.

Suggestions:

- `andEventually`
- `eventually`
- `whenResolved`
- `andThen`

The tricky part to me is that these all could just as well be read as aliases for `map`, since the difference between `map` and `flatMap` comes down to `map` returning a `Deferred<Value>` that's been resolved from the get-go, while `flatMap` allows you to return any flavor of `Deferred<Value>`.

zwaldowski:
I think I'm able to be convinced the `map` and `flatMap` are the wrong names entirely. There's a slight conceptual mismatch with the stdlib (`map` and `flatMap` are used mostly for containers), as well as the signature differences (under 2.0, `rethrows`, etc.).

Let's reframe this suggestion around what we ultimately want the API to look like in terms of clarity and ease-of-use. Would a `Future` type have names `map` and `flatMap`?

zwaldowski:
Ping @jjmanton for his opinions.

jjmanton:
Any thoughts on C++/Boost futures? http://www.boost.org/doc/libs/1_55_0/doc/html/thread/synchronization.html#thread.synchronization.futures.reference.unique_future.then

They use method names like then, upon, unwrap.

jeremy-w:
Some comments in the Boost docs link to [a proposal to improve C++11's std::future/std::shared_future](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2013/n3634.pdf); I found that document did a lot better job of explaining what the APIs were supposed to do and why. The proposal is inspired by Microsoft's Parallel Programming Libraries (PPL).

The overall model is more complicated than Deferred, and more synchronous; that last bit is what the  proposal doc that the Boost API is in the process of implementing is about fixing.

- You get a future either directly (`make_ready_future`) or from either a promise or a task.
- A promise can only be filled once (same as how Deferred works today).
- A future can only be gotten once. Getting it again triggers an exception.
  - You can work around this by making a shared future.
- A get is blocking.

So Future.get is Deferred.value (though .get has a version with a timeout parameter).

The C++ proposal adds API to enable writing non-blocking code with futures by adding:

[...]

Issue #20, opened 2015-08-07
change "fill" names to something more suggestive of the "not-yet-determined value" semantics


jeremy-w:
"Fill" suggests a "box" viewpoint. This doesn't mesh well with the "not-yet-determined value" semantic POV.

Suggestions:

- determine
- resolve
- fix
- now
- settle

Splitting into Promise/Future means that we want a term more in line with the "promise" concept. Suggestions:

- resolve (used by the pasteboard APIs)
- fulfill (suggested by @lyricsboy and keeps the 'fill' feel without committing to a box concept)
- deliver

zwaldowski:
I haven't found people hanging up on `fill`, though there's definitely an argument to be had. I feel `determine` is too wordy, and I don't _quite_ know why I was using it in the docs but for following the "not using the word in the definition" rule.

lyricsboy:
I am fine with `fill` in my use, but `fulfill` might be a better word. It is basically a promise, after all.

zwaldowski:
And in the case of #33 it's literally a `Promise`... `Type`. :P

jeremy-w:
Can we make `Task` into `FingersCrossedPromise` so that you can `FingersCrossedPromise.break` rather than `Task.cancel`? ;)

`fulfill` plays well with the "Promise" verbiage for sure. Though I notice that the pasteboard API uses the term "resolve a promise" rather than "fulfill".

zwaldowski:
`break` is a language keyword, so I think not. :wink:

jeremy-w:
/me tableflips about Swift and its keyword-madness again.

zwaldowski:
Then you might as well tableflip about C too...

zwaldowski:
The most movement on this is `fulfill`. Do we want to commit to that?

jeremy-w:
:-1: `fulfillIfUnfulfilled` is a beast to type. `resolve` is common and rolls off the tongue easier, but no-one seemed interested, so let's leave it as `fill`. It's used enough that brevity matters, too.

zwaldowski:
`IfUnfulfilled` is no longer around after the Pod's decision. But alright!

jeremy-w:
I just cruised through tests on master hunting for uses of `fill` to get a feel for what a more verbose word would do. Sounds like we really need to hit the merge button soon on a lot of things. :)

zwaldowski:
[...]

Issue #19, opened 2015-08-07
rename `fillIfUnfulfilled` to `fill`


jeremy-w:
This is a "gotcha" in the API for now. It's common to have multiple producers for a deferred value (say, both a network operation AND possibly the user smacking a cancel button); having to remember to do `fillIfUnfulfilled` or risk exploding is a wart from my point of view.

zwaldowski:
Ping @randomstep so he sees this conversation, pursuant to [his comment](https://github.com/bignerdranch/Deferred/pull/16#issuecomment-131473569).

zwaldowski:
I'm not going to put words in anyone's mouths, but @jgallagher and I are together in disagreement about this. Common-case usage of `Deferred` itself (that is, not considering `Task` or Future/Promise) cases should assert if filled because it models being filled once-and-only once. In this way (w.r.t. @randomstep's "but don't load the metaphorical crashing gun"), the precondition in `Deferred` is more like Core Data threading assertions: silently eating the error would be obscuring programmer error and burying actual potential problems. Abstractions that truly do require racing behavior should take care to deactivate the assertion if that is correct for their abstraction.

randomstep:
So ideally Deferred would only have a single owner / filler (something like what Rust does with the borrow checker). At least, it would be ideal if the 'once and only once' usage of fill could be enforced, rather than just documented. I seem to remember hearing some discussion of this previously.

So, putting aside the eventual API for a moment, Deferred has 2 uses:
1 - only a single provider will fill the Deferred, once and only once. If anyone else tries that is a semantic error.
2 - first come, first ... filled. Multiple providers can hold a reference to Deferred. Whoever fills it first wins. Everyone must check before filling, since someone else could get there first. Any later fills are just discarded.

`fill` is the current API for 1. `fillIfUnfilled` is the current API for 2. I think API 2 is clear. But these are really 2 completely different usages of Deferred. Would be great if the type system (or at least API) made clear that case 1 was different. I'm guessing that is more easily said than done though ... at a minimum I could imagine `fillOnlyOnce` being the natural corollary rather than `fill`.

randomstep:
What if there was an `OwnedDeferred` that could only be owned by one instance at a time? I guess the OwnedDeferred (or a manager of said instance) would have to have a reference to the owner/filler, which could be updated when the owner wanted to pass ownership.

zwaldowski:
[...]

PR #33, opened 2015-09-06
Initial Swift 2 restructuring

zwaldowski:
* Splits the convenience APIs of `Deferred` out to `FutureType` (the reading/chaining APIs) and `PromiseType` (the writing APIs).
   - This'll greatly ease the creation of an idiomatic `Task`. A module would declare a bunch of protocol extensions on `FutureType where Value: ResultType`, then implement `Task` as a concrete type satisfying that constraint.
   - This is probably not its final form; there's been some discussion of `Deferred` becoming `private` as a result of a real honest-to-god split.
* Modeling after the Swift 2.0 stdlib, introduces an `AnyFuture` for erasing the type of an underlying `Deferred` or `Task`. This effectively solves the anybody-can-fill problem while keeping `Deferred` useful on its own when not crossing API boundaries.
   - One of a few `AnyFuture` specializations adds an `ignoringResult` to any `FutureType`, resolving #24 safely.

This PR does not attempt to make refactors other than moving things out to protocols. Notably, this means the documentation to `Deferred` was lifted directly; the references to a "deferred value" may not make sense. However, this PR would serve as a great battleground for issues #18, #19, #20, and #21.


jeremy-w:
What is the use case for the `AnyFuture`? It looks like the `Any` types in the Swift stdlib are somewhat far afield if you're still mostly integrating with Cocoa libraries.

zwaldowski:
I don't think that assessment makes much sense. Aren't there all sorts of things we use in Swift that aren't Cocoa-compatible? We use them anyway because we're using them from Swift.

Type erasure, as applied here, solves the problem of having protocol types when the protocol has associated types; the design of protocols is currently is opposed to being able to return a `CollectionType<T>` from something.

Consider a public type that privately stores information in a `[Int: String]`. You might want to perform some operation on the keys of that dictionary:

```swift
    let decremented = aDict.keys.filter { $0 > 5 }.map { $0 - 5 }
```

If you wanted to move that out to an API on its own, the type of `decremented` is `LazyMapCollection<LazyFilterCollection<LazyMapCollection<Dictionary<Int, String>, Int>>, Int>`. Not easy to swallow. Worse, if some user of the API makes assumptions about that type, it'll break if you change the map/filter later. Instead of reifying into `[Int]` for something that's going to be enumerated once and thrown away, `AnySequence<Int>` is simpler than exposing the full type and less expensive than creating new array storage.

`AnyFuture` works similarly, but exists to solve the problem of being able to fill a `Deferred` or `Task` at any time, even if its intended use by the author of an API is effectively read-only. Having concrete types that implement `FutureType` to solve these problems or introduce optimizations would introduce coupling in users' code (like with the crazy signature above).

Some concrete use cases of `AnyFuture` like this:

* A library like CoreDataStack may want to use a Future-based control flow without exposing the specific flavor of `Deferred` it uses across the API boundary.
* Store types in an app might expose a `AnyFuture<Void>` (using `ignoringResult`) to represent its current operation, so the app will not attempt to log out until all operations are done. This would prevent an automated, un-cancellable sync with Core Data from causing exceptions because the store was deleted out from under it.
* Depending on some logic, for instance, a cache hit, a piece of code might want to use `Deferred` sometimes and `Task` others, but the rest of the code only cares that it's getting a `UIImage` back at some point.

jeremy-w:
The assessment was more "this isn't immediately obvious because those types aren't necessarily in your face", not "we don't need this because it's Crazy Swift". :)

I should have just asked for a doc comment, sorry. Anyway, everything in your response from "`AnyFuture` works similarly…" on would make a great doc comment on `AnyFuture`!

jeremy-w:
> the problem of having protocol types when the protocol has associated types

Just to make sure I'm following: This is the dreaded "error: protocol 'P' can only be used as a generic constraint because it has Self or associated type requirements", right? So this is simultaneously a hack around that annoyance as well as a way to prevent unnecessary concrete classes / types that expose implementation details (your lazyXYZCollection example).

jeremy-w:
Found the link in the file-level comments to https://realm.io/news/type-erased-wrappers-in-swift/, which goes a ways towards explaining what's going on in general. Its link to the Wikipedia article on type erasure is unfortunately worse than useless (the article addresses type-erasure in the sense of an operational semantics that does not depend on types, like how Java generics work), and the conclusion hand-waves about why its AnySequence needs to be a class by pointing out that Apple's version is also a class, which is not actually correct – AnyGenerator is a class, yes, but AnySequence is a struct.

zwaldowski:
As I commented above, yeah, I thought I had written it down as documentation. I really appreciate the critical eye.

Re: "simultaneously", yes, you're right. It covers two purposes.

Re: the article, it's not wrong, just a few weeks out of date. They keep changing around the implementations. Mine is based around the universe at it is in Beta 6. I do agree about the hand-waving.

There should be a more concrete explanation (if not in the docs, maybe a wiki/FAQ/blog post or something) about what's happening. I agree with the hand-waving, and the term "type erasure" is not completely useful on its own here.

jeremy-w:
I think [this](http://stackoverflow.com/a/6324714/72508) explains AnyGenerator and friends far more directly (and less verbosely) than does [this](https://realm.io/news/type-erased-wrappers-in-swift/), but I can’t tell if it would be too brief if I hadn’t read the latter first.

Unsurprisingly, it's a discussion of the same trick as used in C++.

zwaldowski:
Updated to clean up branch history and fix Xcode project issues.

zwaldowski:
Resolved dependency on #30.

Deferred/FutureCollections.swift
```diff
@@ -0,0 +1,53 @@
...
+//
+
+import Dispatch
+
+public extension SequenceType where Generator.Element: FutureType {
+    /// Choose the future that is determined first from a collection of futures.
+    ///
+    /// - returns: A deferred value that is determined with the first of the
+    ///   given futures to be determined.
+    var firstFuture: Deferred<Generator.Element.Value> {
```
zwaldowski: Having `Future` in the name of the method isn't discoverable. Something like `firstFinished`?

Deferred/FutureCollections.swift
```diff
@@ -0,0 +1,53 @@
...
+        return combined
+    }
+}
+
+public extension CollectionType where Generator.Element: FutureType {
+    /// Compose a number of futures into a single deferred array.
+    ///
+    /// - returns: A deferred array that is determined once all the given values
+    ///   are determined, in the same order.
+    var allFutures: Deferred<[Generator.Element.Value]> {
```
zwaldowski: Same as with `firstFuture`. `allFinished`?

Deferred/FutureCollections.swift
jeremy-w: Verbiage in documentation throughout uses `determined`, so maybe `firstDetermined`? This mirrors the docstring on this method as well.

Deferred/FutureCollections.swift
jeremy-w: `asFutureArray`? It's a transform from array-of-futures to future-array, roughly `[Future<T>] --> Future<[T]>`.

Deferred/FutureCollections.swift
zwaldowski: Good thinking. Of course, we're also not 100% on board with the verbosity of `determine` (#20).

jeremy-w: Might be better to use "earliest" rather than "first", since that gets the time aspect in there directly, while the latter is more indirect "if we map time onto a timeline, then the earliest one is the first one as you move forward".

Deferred/FutureCollections.swift
zwaldowski: Trying to avoid using `future` in the names of these, such that you don't have to understand the whole protocol system to get by with just `Deferred`

jeremy-w: `determinedValues`? `joinPoint` / `joinedValue` (harks back to the fork/join description I read in the docstring on FutureType IIRC)? `allTogether`?

Deferred/FutureCollections.swift
jeremy-w: #20 doesn't include any discussion of verbosity problems. I think you have a mental draft that needs posting. ;)

zwaldowski: I usually do… :P

wolfbrain:
:eyes: @jeremy

zwaldowski:
Updated for new `fill` semantics after significant discussion.

zwaldowski:
Went ahead and made some of those naming changes.

Deferred/AnyFuture.swift
```diff
@@ -0,0 +1,113 @@
...
+/// `Value` type, hiding the specifics of the underlying `FutureType`.
+///
+/// Authors can use this type to:
+///
+/// - Prevent clients from coupling to the specific kind of `FutureType` your
+///   implementation is currently using.
+/// - Publicly expose only the `FutureType` aspect of a deferred value,
+///   ensuring that only your implementation can fill the deferred value
+///   using the `PromiseType` aspect.
+public struct AnyFuture<Value>: FutureType {
```
zwaldowski: I wonder if  should go ahead and name this `Future`. Yeah, sure, it's out of sync with the Swift standard library, but it's essentially what we want.

PR #38, opened 2015-10-25
Throwing `fill`

### LINKED ISSUE #19: rename `fillIfUnfulfilled` to `fill`

zwaldowski:
Depends on #33.

In light of the return of synchronous fill, `fill` can be made into a throwing API. This would neatly resolve #19.

The old behavior continues to exist as an overload, `PromiseType.fill(_:assertIfFilled:)`.

Pros:
* No unexpected behavior when filling a `Deferred` twice
* Does not obscure programmer error
* Implementers of `PromiseType` are free to define their own errors
* Programmer makes choice/is aware if they are encouraging races or not, by picking `try?` or `try!`

Cons:
* There's no really effective way to "handle" an error
* Common case use would litter your code with `try!`


zwaldowski:
Updated to clean up branch history.

Deferred/FutureType.swift
```diff
@@ -0,0 +1,164 @@
...
+    /// - parameter queue: Optional dispatch queue for starting the new
+    ///   operation from. Defaults to a global queue matching the current QoS.
+    /// - parameter transform: Start a new operation using the deferred value.
+    /// - returns: The new deferred value returned by the `transform`.
+    /// - seealso: Deferred
+    func flatMap<NewFuture: FutureType>(upon queue: dispatch_queue_t = genericQueue, transform: Value -> NewFuture) -> Deferred<NewFuture.Value> {
+        let d = Deferred<NewFuture.Value>()
+        upon(queue) {
+            transform($0).upon(queue) {
+                d.fill($0, assertIfFilled: true)
```
randomstep: Once we transform this Deferred (or just in general any time we pass around a Deferred), anyone can fill it. Which means I can write a nice safe API, fill only once (with crash if I mess up) - and then someone else can come along and fill with the Deferred I gave them, leading to a crash. Ideally I could hand out a Deferred (or something else) that they could only read, not write to (fill).

zwaldowski: This is technically already true, and always has been! :worried: It's what I call the "anybody can fill a `Deferred` problem" (which I learned today I never filed an issue for!). That can be solved by the type-system in #33 if `map` and `flatMap` returned `AnyFuture`, or by changing these to `assertIfFilled: false`.

zwaldowski: Kinda relevant to #19. And I'm sure @jeremy-w will be interested.

jeremy-w: Yes, this has been under discussion for a while. I advocated fill not exploding due to how I understand the semantics of `Deferred`. Removing an independent `Deferred` in favor of a `Future` (read) and `Promise` (write) pair routes around that question and is a lot clearer, at the cost of having to keep track of two things rather than just one in client code.

zwaldowski:
Closing because, oh, God, the **screaming**.

PR #41, opened 2015-10-27
Updating the comment as Deferred is not subscriptable.

hitsvilleusa:


zwaldowski:
We should consider rephrasing the comment anyway, as `someDeferred.value[i]` is not something you should typically do. I think someone said something to that effect in Slack. Anyway, merging. :+1:

PR #42, opened 2015-10-27
Removes the nested nature of the Deferreds with any

hitsvilleusa:
It seems that when using `any`, you are saying ... I don't care about which one comes back.  Tracking the `Deferred<Deferred<T>>` seems like overkill since I didn't care about the specific one.  If I do care about the specific one, then I could likely track it using a tuple or something more robust in the Result.

Working with the nested Deferreds seems to add more confusion than benefit, and as it stands now, it is more in parallel with the `all` function.


zwaldowski:
Fantastic, thanks! I'll incorporate this into #33. :+1:

Issue #54, opened 2015-12-08
Linux support


zwaldowski:
This is a tracking ticket for improvements related to multi-platform support.

- [x] `#if os`-out use of NSProgress
- [x] Set up CI

jeremy-w:
I think the first step towards replacing GCD might be pushing it on the other side of a minimal API we control. Then replacing it is a matter of finding something else that can replace the API, possibly on a per-platform basis (i.e., we could still use GCD when running under OS X).

zwaldowski:
I have a draft of protocol-ifying the user of `dispatch_block_t`, but I'm not sure how to start abstracting queues. There's nothing that comes in an average Linux base install to do work queues. The closest common-use thing I can think of at the high-level is [libuv](https://github.com/libuv/libuv). Or there's stuff at the low-level like libpthread_workqueue, which libdispatch is going to depend on. In either case, for the work involved in a dependency like that I'd almost rather wait for the first-pass implementation of GCD queues.

zwaldowski:
Bridging in a C library also has complications in the lack of block support, which we'd have to write ourselves.

jeremy-w:
Are we leaning on synchronous queues anywhere? If we're mainly using queues as target execution contexts, then reworking in terms of an "Executor" concept might suffice.

zwaldowski:
That sounds reasonable. What would an `Executor` do in a non-queueing system? Would it just be a thread target?

Other challenge posed: the `CallbacksList` (let's call it; today, the `dispatch_block_t`) is linked to the implementation of `Executor` (today `dispatch_queue_t`) through the use of `dispatch_block_notify`, which evinces a significant performance win. I don't want component swapping to lose that design as a matter of course.

zwaldowski:
I've already been pondering `Executor` outside of cross-platform support; there are other things you might have in your Cocoa API that don't use dispatch queues but want to "perform" on (`NSOperationQueue`, `NSManagedObjectContext.performBlock(_:)`). Those would also lose the benefit of the `_notify` concepts.

zwaldowski:
Updated task list with discussion from above.

zwaldowski:
Looks like GCD is [impending](https://github.com/apple/swift/commit/06860a4d0102db0cd9dfd87c7d0a3fcb9939ef91), so with #69 we might be able to resolve this issue. If we think `Executor` is important for things-that-are-queues-but-`dispatch_queue_t`, we should split that out into a different issue.

[...]

PR #57, opened 2015-12-09
DeferredBuffer cleanups

zwaldowski:
* Extricates use of `dispatch_block_t` scattered throughout `Deferred` and `DeferredBuffer` into a protocol-backed `CallbacksList` - part of #54.
* Refactors `DeferredBuffer` out into a different type for readability.
* Following Swift's [collections](https://github.com/apple/swift/blob/0b0123f39e9202e9a7adc2a1f242134720f9a79c/stdlib/public/core/Arrays.swift.gyb), uses `ManagedBufferPointer` over `ManagedBuffer`. Loses a little magic to be a little easier to read.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,6 +18,7 @@ xcuserdata
 *.xccheckout
 *.moved-aside
 *.xcuserstate
+*.xcscmblueprint
 
 ## Obj-C/Swift specific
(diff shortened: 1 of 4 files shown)


randomstep:
:+1:

Issue #55, opened 2015-12-08
Require `LockProtected` members to be `rethrows`.


zwaldowski:
Instead of `withReadLock<Return>(@noescape body: Value -> Return) -> Return`, we should have `withReadLock<Return>(@noescape body: Value -> throws Return) rethrows -> Return` so you can throw across lock boundaries. Same with `withWriteLock`.

PR #58, opened 2015-12-09
Convert all ReadWriteLock, LockProtected to rethrows

zwaldowski:
Addresses #55


randomstep:
All good. Nice use of defer.

PR #60, opened 2015-12-09
Rename AnyFuture, elevate IgnoringFuture

zwaldowski:
I could've sworn I filed an issue for this earlier, but it seems I never submitted it.

* Since it's mostly the API we already want, rename `AnyFuture` to just `Future`. If we want to have `Future` be a simpler implementation that owns something private and mutable, and just kill `Deferred` entirely, this provides forward coverage for that.
* Marking `FutureBoxBase` as anything but `private` is a code smell. Like `FlattenSequence`, `EnumerateSequence`, etc. in the stdlib, `IgnoringFuture` should wrap the type directly and not care about type erasure.


randomstep:
Looks good.

PR #61, opened 2015-12-09
Platform atomic agnosticism 

zwaldowski:
Depends on #57 and #58.

Contributes to #54:
* Reshuffle project format for Swift Package Manager
* Adopt C11 atomics when available using bignerdranch/AtomicSwift.


Sources/Deferred/ReadWriteLock.swift
```diff
@@ -0,0 +1,258 @@
...
+            // decrement reader count
+            repeat {
+                let state = __bnr_atomic_load_32(_state)
+                let readerCount = state & Masks.MASK_READER_BITS
+                
+                // sanity check that we have a positive reader count before decrementing it
+                assert(readerCount > 0, "unlocking read lock - invalid reader count")
+                
+                // desired new state: 1 fewer reader, preserving whether or not there is a writer waiting
+                let newState = (readerCount - 1) | (state & Masks.WRITER_WAITING_BIT)
```
jeremy-w: For future-proofing, might be better to just mask out existing reader portion, vs masking in what is (currently) the only other field?

```
let newState = (readerCount - 1) | (state &~ Masks.MASK_READER_BITS)
```

zwaldowski: I don't anticipate more masks, the whole point is readers-writer in a single word. @jgallagher might be able to speak to it better, he adapted the original implementation; I'm just here to add underscores to the function calls.

zwaldowski: Seems GitHub has misidentified the file move + changes as a new file. :disappointed: Odd, because Git doesn't for the changes. Anyway, [the original implementation](https://github.com/bignerdranch/Deferred/pull/61/files#diff-1311f678fc962b574d7935db66cb98faL133).

zwaldowski:
Putting this PR on ice while I decide what to do about spinlock priority inversion.

Issue #67, opened 2016-01-07
peek(), isFilled return nil/false after fill()


preble:
The following test demonstrates the issue (should be added to ExistentialFutureTests):

```
func testFill() {
    let deferred = Deferred<Int>()
    anyFuture = Future(deferred)
    XCTAssertFalse(anyFuture.isFilled)               // pass
    deferred.fill(42)
    XCTAssertNotNil(anyFuture.peek())                // fail
    XCTAssertTrue(anyFuture.isFilled)                // fail
    XCTAssertNotNil(anyFuture.wait(.Now))            // fail
    XCTAssertNotNil(anyFuture.wait(.Interval(0.1)))  // pass
}
```

The problem can also be observed on a pure Deferred (not wrapped in Future).

In Deferred.wait, it appears that dispatch_block_wait() returns timeout/error when called with a timeout of zero/now, as the "upon" block (which assigns `value = $0`) is only queued and has't completed.

preble:
I should add that this is also observable in a more typical async setup by adding similar XCTAsserts to the end of ExistentialFutureTests.testAnyWaitWithTimeout().

preble:
Thanks, @zwaldowski!

PR #68, opened 2016-01-07
Fixes Deferred.wait(_:) when filled

zwaldowski:
Addresses #67.

### DIFF
--- a/Sources/Deferred.swift
+++ b/Sources/Deferred.swift
@@ -91,18 +91,26 @@ public struct Deferred<Value>: FutureType, PromiseType {
     :returns: The determined value, if filled within the timeout, or `nil`.
     */
     public func wait(time: Timeout) -> Value? {
-        let queue = dispatch_get_global_queue(QOS_CLASS_UTILITY, 0)
-        var value: Value?
-        let handler = upon(queue, options: DISPATCH_BLOCK_ENFORCE_QOS_CLASS) {
-            value = $0
+        var result: Value?
+        func assign(value: Value) {
+            result = value
         }
+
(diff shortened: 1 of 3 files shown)


Sources/Deferred.swift
```diff
@@ -91,18 +91,26 @@ public struct Deferred<Value>: FutureType, PromiseType {
...
-        let queue = dispatch_get_global_queue(QOS_CLASS_UTILITY, 0)
-        var value: Value?
-        let handler = upon(queue, options: DISPATCH_BLOCK_ENFORCE_QOS_CLASS) {
-            value = $0
+        var result: Value?
+        func assign(value: Value) {
+            result = value
+        }
+
+        // Pure FutureType implementations implement isFilled in terms of wait()
```
jgallagher: This comment implies the next line would be infinitely circular, wouldn't it? If wait calls isFilled and isFilled is implemented in terms of wait.

Sources/Deferred.swift
```diff
@@ -91,18 +91,26 @@ public struct Deferred<Value>: FutureType, PromiseType {
...
+        if case .Now = time where isFilled {
```
jgallagher: Why only do this for `.Now` and not all timeouts?

zwaldowski: …fair.

Sources/Deferred.swift
zwaldowski: No, we implement `isFilled` ourselves. "Pure" meant as "Only" (i.e., not `PromiseType`). Will clarify.

zwaldowski:
Comments addressed.

jgallagher:
:+1:

PR #69, opened 2016-01-25
Atomics shimming

zwaldowski:
Working toward #54 by shimming `OSAtomic` when running on not-Darwin using the [AtomicSwift](https://github.com/bignerdranch/AtomicSwift) header. Also adds an extra shim, `_OSAtomicSpin`, to both versions, which allow hyperthreads to continue while livelocking and improve measurable speeds.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 ## Build generated
+.build/
 build/
 DerivedData
 docs/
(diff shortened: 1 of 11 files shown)


zwaldowski:
Also adds initial support for SwiftPM! :clap:

PR #74, opened 2016-02-18
Arguments label API fixes

zwaldowski:


randomstep:
:+1: I like this change.

randomstep:
Any documentation or example updates needed?

zwaldowski:
@randomstep: No, doc comments should use the internal variable name, so it stays "transform". If we're using the full signature anywhere, i.e., `map(upon:transform:)` should become `map(upon:_:)`, so I'll do a brief look for those.

zwaldowski:
Nope, there are none. I take it you're gonna make me merge it myself? :stuck_out_tongue:

Issue #64, opened 2015-12-19
Task


zwaldowski:
This is a tracking issue for what I imagine to be several PRs. Work is ongoing at branch [zwaldowski/task](https://github.com/bignerdranch/Deferred/tree/zwaldowski/task) and [bignerdranch/Task](https://github.com/bignerdranch/Task). Vomiting my plan out on the page because I'm taking vacation. :sunglasses: :ship:

Deferred is very often used to express the result of operations that can naturally fail, expressed rationally as `Deferred<Result<T>>`. Deferred should include extensions to make this easier.

- [x] Add an extremely barebones `Result` and `ResultType`.
  - [x] Full test coverage.
- [x] Subtree merge a cleaned up version of the work at [bignerdranch/Task](https://github.com/bignerdranch/Task) (private repo).
- [x] Tree-filter to move `Task/**/*.swift` to `Sources/Task`.
- [x] Tree-filter to move `Task/TaskTests/**/*.swift` to `Tests`.
- [x] Implement (or finish implementing) Result primitives as `FutureType where Value: ResultType` extensions.
  - [x] `uponSuccess` and `uponFailure`
  - [x] `map` and `flatMap` overloads (or `mapSuccess` and `flatMapSuccess`)
     - Should `flatMap` versions throw instead of returning `Result`?
  - [x] `recover` and `recoverWith` (or `mapFailure` and `flatMapFailure`)
  - [x] async convenience initializers (`init(_: () -> T)`, `init(_: () throws -> T)` for `Result` versions)
  - [x] `flatten`
- [ ] Eating our vegetables.
  - [x] Documentation coverage for new code.
  - [ ] As-high-as-possible test coverage.

zwaldowski:
Closing in favor of #98.

PR #75, opened 2016-02-19
Lazy map and flatMap

zwaldowski:

* Adds a `lazy` subtree of types, derived from Swift's `LazyCollectionType`, on which `map` and `flatMap` can be modeled lazily.
  - `lazy.map(_:)` executes the transform on each call to `upon` or `lazy`. This is a semantics and performance improvement for simple conversions.
  - `lazy.flatMap(_:)` defers the start of the new operation until a call to `upon` or `lazy` is made.
  - The types introduced will reduce code duplication needed in #64.
* Introduces a `flatten()` method (incl. a lazy definition) for combining two `Future`s, if you are ever so unfortunate as to get one of those.
* Introduces a `ignore()` method to replace `IgnoringFuture`.

* [ ] Might be able to wrangle a performance improvement from `and` by `lazy.map`-ing to `Any` and using `joinedValues`. Need to do tests; should attach to a separate issue.

#### Testing

* [ ] Add tests for the new guarantees made by `lazy`.

#### API Changes

The `lazy` constructs are additive. The little-used `IgnoringFuture` type is replaced by the new lazy getters. Though `IgnoringFuture` is removed, its symbol remains with migration guidance.

`IgnoringFuture` would be removed in the final release of 2.0.


zwaldowski:
Depends on #83.

zwaldowski:
Also depends #86.

zwaldowski:
Split unrelated improvement out into #87.

zwaldowski:
Bumping this to 2.1 because of the mostly additive nature. I'll backport `ignore()` shortly.

zwaldowski:
Oh, and the `flatMap` optimization.

zwaldowski:
Will reopen later

Issue #65, opened 2015-12-19
Dis-integrate "docs.md"


zwaldowski:
The legacy [`docs.md`](https://github.com/bignerdranch/Deferred/blob/master/docs.md) contains some high-level documentation and usage examples. The high-level stuff should be pulled into the README (if there is any mismatch; the README is already comprehensive). The usage examples should be moved into the HeaderDoc.

Issue #32, opened 2015-09-05
doc.md refers to method deprecated in 2.0


mdmathias:
Should replace `toInt()` with `Int`'s failable initializer.

PR #78, opened 2016-03-20
Merged relevant material from docs.md into the README.md

davidahouse:
This PR addresses issues #32 and #65.


README.md
```diff
@@ -210,9 +213,125 @@ func AsynchronousFriends(forUser: jimbob) -> Deferred<[Friend]> {
...
+## Usage - Producer
+
+```swift
+// Potentially long-running operation.
+func performOperation() -> Deferred<Int> {
+		// 1. Create deferred.
+    let deferred = Deferred<Int>()
+
+		// 2. Kick off asynchronous code that will eventually...
+    dispatch_async(dispatch_get_main_queue(), {
```
zwaldowski: Change to a background queue; we want to demonstrate the thread safety aspect.

README.md
```diff
@@ -210,9 +213,125 @@ func AsynchronousFriends(forUser: jimbob) -> Deferred<[Friend]> {
...
+}
+```
+
+### Blocking on Fulfillment
+
+Use the `value` property to wait for the `Deferred` to be filled and get the value.
+
+```swift
+// WARNING: Blocks the calling thread!
+let result: Int = performOperation().value
```
zwaldowski: We removed `.value` as a public API. We could demonstrate the use of `wait(_:)` instead.

README.md
```diff
@@ -210,9 +213,125 @@ func AsynchronousFriends(forUser: jimbob) -> Deferred<[Friend]> {
...
+```
+
+### Chaining Deferreds
+
+Monadic `bind` and `map` are available to chain `Deferred` results. For example,
```
zwaldowski: `flatMap` now

README.md
```diff
@@ -210,9 +213,125 @@ func AsynchronousFriends(forUser: jimbob) -> Deferred<[Friend]> {
...
+suppose you have a method that asynchronously reads a string, and you want to
+call `toInt()` on that string:
```
zwaldowski: `String.init(_:)` now

README.md
```diff
@@ -210,9 +213,125 @@ func AsynchronousFriends(forUser: jimbob) -> Deferred<[Friend]> {
...
+		return deferredResult
+}
+
+// Consumer
+let deferredInt: Deferred<Int?> = readString().map { Int($0) }
+```
+
+`bind` and `map`, like `upon`, execute on a concurrent background thread by
+default (once the instance has been filled), unless a different queue is
+passed when the `Deferred` instance is created. `bindQueue` and `mapQueue` are
```
zwaldowski: These are `flatMap(upon:_:)` and `map(upon:_:)` now. `upon` has a default parameter

README.md
```diff
@@ -210,9 +213,125 @@ func AsynchronousFriends(forUser: jimbob) -> Deferred<[Friend]> {
...
+
+### Combining Deferreds
+
+There are three functions available for combining multiple `Deferred` instances:
+
+```swift
+// `both` creates a new Deferred that is filled once both inputs are available
+let d1: Deferred<Int> = ...
+let d2: Deferred<String> = ...
+let dBoth : Deferred<(Int,String)> = d1.both(d2)
```
zwaldowski: `and(_:)` now

README.md
```diff
@@ -210,9 +213,125 @@ func AsynchronousFriends(forUser: jimbob) -> Deferred<[Friend]> {
...
+
+// `all` creates a new Deferred that is filled once all inputs are available.
+// All of the input Deferreds must contain the same type.
+var deferreds: [Deferred<Int>] = []
+for i in 0 ..< 10 {
+		deferreds.append(...)
+}
+var allDeferreds: Deferred<[Int]> = all(deferreds)
```
zwaldowski: `let` and `. joinedValues` now.

README.md
```diff
@@ -210,9 +213,125 @@ func AsynchronousFriends(forUser: jimbob) -> Deferred<[Friend]> {
...
+// Once all 10 input deferreds are filled, allDeferreds.value[i] will contain the result
+// of deferreds[i].value.
+
+// `any` creates a new Deferred that is filled once any one of its inputs is available.
+// If multiple inputs become available simultaneously, no guarantee is made about which
+// will be selected.
+var anyDeferred: Deferred<Deferred<Int>> = any(deferreds)
```
zwaldowski: The use of nested `Deferred` has been replaced with `earliestFilled` on a sequence of futures.

README.md
```diff
@@ -210,9 +213,125 @@ func AsynchronousFriends(forUser: jimbob) -> Deferred<[Friend]> {
...
+### <a name="upon"></a>Running Closures Upon Fulfillment
+
+You can use the `upon` method to run a closure once the `Deferred` has been
+filled. `upon` can be called multiple times, and the closures will be called
+in the order they were supplied to `upon` (with the normal race condition caveat
+if you are calling `upon` from multiple threads simultaneously).
+
+By default, `upon` will run the closures on a background concurrent GCD queue.
+You can change this by passing a different default queue when the `Deferred` is
+created, or by using the `uponQueue` method to specify a queue for the closure.
```
zwaldowski: `uponQueue` is now `upon(_:body:)`

README.md
davidahouse: Thanks for the comments! working on updates.

README.md
```diff
@@ -210,9 +213,125 @@ func AsynchronousFriends(forUser: jimbob) -> Deferred<[Friend]> {
 }
 ```
 
+## Usage - Producer
```
jeremy-w: Recommend an h2 "Usage" with h3 "Producer" and "Consumer" below, rather than prefixing two h2s with Usage, to clarify the structure.

Even better: steal the "Tasks" concept from Apple's docs and be a bit more descriptive than just "Producer" (of what?), maybe something like:

```
## Tasks
### Vending a Future
### Taking Action when a Future Is Filled
### Acting when First of Several Futures Gets Filled
### Acting when All of Several Futures Have Been Filled
```

and so on.

jeremy-w:
This README might have reached the point where a Table of Contents is appropriate. I generally use `doctoc` for that (npm -g install doctoc).

zwaldowski:
@jeremy-w To that point, the TOC'ed and usage-based readme should probably be transcluded in Jazzy, as if it were the "Deferred Type Programming Guide" (as long as we're going for Apple-like).

zwaldowski:
Totally outside the scope of this PR, but just for reference I had in mind what [ReSwift](http://reswift.github.io/ReSwift/master/) does — it's sort of the poster child for Jazzy. Notice that they split up their README into different Markdown documents that get used in the documentation set.

zwaldowski:
Moved ToC discussion into #84.

PR #80, opened 2016-03-23
Improve debugging through reflection support

zwaldowski:
This achieves two primary benefits:

* Futures (i.e., `Deferred`) and `LockProtected` have sidebar results in playgrounds through `Mirror`.
* Futures (i.e., `Deferred`, and notably `Future<T>`) and `LockProtected` have better results in the debugger through `debugDescription`.
   - Though Swift has fallback behavior to recover, we have `debugDescription` but not `description` because these descriptions are not intended to be user-facing at any point.

## API Changes

Purely additive. `FutureType` gains inheritance of Swift's reflection protocols, but provides default implementations.

### DIFF
--- a/Sources/ExistentialFuture.swift
+++ b/Sources/ExistentialFuture.swift
@@ -106,4 +106,9 @@ public struct Future<Value>: FutureType {
     public func wait(time: Timeout) -> Value? {
         return box.wait(time)
     }
+
+    /// Return the `Mirror` for `self`.
+    public func customMirror() -> Mirror {
+        return Mirror(reflecting: box)
+    }
 }
--- a/Sources/FutureType.swift
+++ b/Sources/FutureType.swift
@@ -198,24 +198,14 @@ public extension FutureType {
     }
 }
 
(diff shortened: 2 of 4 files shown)


randomstep:
:+1: for code changes, haven't tested in practice though.

On `description`: it isn't end-user text (at least not necessarily). "This textual representation is used when values are written to an output stream, for example, by print." 

I tried to find further info from sources from Swift mailing list but didn't yet. But what I took away from discussions I've seen there: `description` is a short representation of the type itself, whereas `debugDescription` is a more verbose textual output intended to help with debugging, often including more info about children or otherwise fuller (and larger) set of data.

zwaldowski:
Perhaps I misspoke in saying "user-facing", but, regardless, supporting `description` seemed not to make sense, in line with `COpaquePointer`, `CGPoint`, `CGSize`, `CGRect`, `Optional`, the pointer types, and the vector types.

zwaldowski:
I've given a common `(isFilled: <false/true>)` reflection representation to all `FutureType`s when specialized over `Void`. This should prevent the need for any custom overrides for reflection.

PR #82, opened 2016-03-27
Add a plain, never-filling `Future.init()`

zwaldowski:
The use case is simple: a codebase using `Future<T>` will occasionally need placeholder instances, if nothing else than for providing a [dummy](https://cloud.githubusercontent.com/assets/170812/14065005/a9bf5c74-f435-11e5-96ae-0676fe92734b.png) value for a return type that requires it. Either way, it may be useful to define "a future whose value will never arrive" as a concrete initializer, rather than wrapping a `Deferred` that you never intend to fill (`let f = Future<Foo>(Deferred())`).

## API Changes

Purely additive.

### DIFF
--- a/Sources/ExistentialFuture.swift
+++ b/Sources/ExistentialFuture.swift
@@ -61,6 +61,17 @@ private final class FilledFuture<Value>: FutureBase<Value> {
     }
 }
 
+// Concrete future wrapper that will never get filled.
+private final class UnfilledFuture<Value>: FutureBase<Value> {
+    override init() {}
+
+    override func upon(queue: dispatch_queue_t, body: Value -> ()) {}
+
+    override func wait(time: Timeout) -> Value? {
+        return nil
+    }
+}
+
 /// A type-erased wrapper over any future.
 ///
 /// Forwards operations to an arbitrary underlying future having the same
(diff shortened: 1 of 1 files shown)


Sources/ExistentialFuture.swift
```diff
@@ -61,6 +61,17 @@ private final class FilledFuture<Value>: FutureBase<Value> {
     }
 }
 
+// Concrete future wrapper that will never get filled.
+private final class UnfilledFuture<Value>: FutureBase<Value> {
```
jeremy-w: What do you think about calling this `Never<Value>`? `UnfilledFuture` reads to me like the door is open for it to be a view onto a `Future` that might one day be filled, and just that that wouldn't be visible through the `UnfilledFuture` view. `Never<Value>` seems straight-up "yeah, this'll never happen". As a bonus, it's a simple word rather than a compound. :)

jeremy-w:
@zwaldowski  What's that dummy image link from? It's the clearest presentation I've yet seen of the test double taxonomy.

Sources/ExistentialFuture.swift
zwaldowski: Not unreasonable, since it's private. Will probably do the same to `FilledFuture` - `Box`.

zwaldowski:
@jeremy-w Addressing offline.

PR #87, opened 2016-03-29
Restore protocol default for `FutureType.wait(_:)`

zwaldowski:

Restores the implementation of `wait(_:)` from Deferred 1.0 as a default implementation on `FutureType`.

I can see a situation where someone might be implementing a custom `FutureType` (such as a custom `LazyFutureType` algorithm, or making a Cocoa type be a Swift `FutureType`) and would only *really* want to implement `upon`; this allows that.

#### Testing

- [ ] Add a stub `FutureType` that isn't Deferred to exercise both the fast- and slow-paths to this method.

#### API Changes

Purely additive.


zwaldowski:
Nah, needs #88. Never mind.

Issue #88, opened 2016-03-29
Decouple from `dispatch_queue_t`


zwaldowski:
Splitting off from #54:

>   * Make use of `dispatch_queue_t` generic or pluggable in some way. (`Executor` concept)

This aims to solve two problems: "What if *I* don't have a dispatch queue?" and "What if *the platform* doesn't have dispatch queues?"

For the former, it would be advantageous to have `upon` arguments take something queue-like, rather than a literal `dispatch_queue_t`. In Cocoa, these may be `NSOperationQueue`, `NSManagedObjectContext`, etc.

For the later, after briefly discussing with @jgallagher, we want to keep the conceptual model simple; rather than guarding `upon` availability to not-Linux, or to have more complex routines (like spinning up `pthread`s), we're going to wait for general Dispatch availability, since that'll be in the Swift 3.0/Deferred 3.0 time range anyway. This means that #54 is officially on ice for now.

#### API effects

API changes in this realm should be backwards-compatible.

PR #89, opened 2016-04-07
Decouple from dispatch_queue_t with ExecutorType

zwaldowski:

Addresses #88 through the introduction of an `ExecutorType` protocol. The concept itself still requires GCD, though, so no progress towards #54 yet. (All that can mostly be guarded out with `#if os` at this point, but that step has not been taken yet.)

#### Testing

The tests continue to use the queue-based versions of `upon`. Extra tests should be written to handle the minor new code path in `Deferred` itself.

#### API Changes

This is a source-compatible API change.


jeremy-w:
Looks like a non-GCD `Executor` could unconditionally vend a `nil` `underlyingQueue`. That doesn't solve how tightly wedded `Deferred` is to the `dispatch_block` API, and the ARC-hooking that goes on there would make shimming at that layer a pain IIRC.

zwaldowski:
Yep, that is indeed the case on both counts. (re: `nil` `underlyingQueue`, that's the [default](https://github.com/bignerdranch/Deferred/pull/89/files#diff-b60e11ff6f5b8cff2acbabd90cf70937R19).) This was intended not so much to move away from GCD, but to allow simpler `upon` syntax when you don't have a queue handy. Future plumbing changes can happen post-2.0.

Sources/Deferred.swift
```diff
@@ -28,9 +28,15 @@ private struct DispatchBlockMarker: CallbacksList {
         // Executing the block "unblocks" it, calling all the `_notify` blocks
         block()
     }
-    
-    func notify(upon queue: dispatch_queue_t, body: dispatch_block_t) {
-        dispatch_block_notify(block, queue, body)
+
+    func notify(upon executor: ExecutorType, body: dispatch_block_t) {
```
jeremy-w: "Notify upon queue" made sense. I'm not sure "notify upon executor" does. I think `notify(executor: …` makes more sense; since the argument is an `Executor`, I don't know that a preposition is even needed. It's like "go tell the executor…".

The "body" argument name doesn't make much sense in either case I think. :\

Sources/Deferred.swift
```diff
@@ -56,32 +62,27 @@ public struct Deferred<Value>: FutureType, PromiseType {
 
     // MARK: FutureType
 
-    private func upon(queue: dispatch_queue_t, options inOptions: dispatch_block_flags_t, body: Value -> Void) -> dispatch_block_t {
+    private func upon(executor: ExecutorType, options inOptions: dispatch_block_flags_t, body: Value -> Void) -> dispatch_block_t {
```
jeremy-w: With the new "executor" verbiage, "body" might make more sense as "execute body: …".

jeremy-w: I wonder if a replacement for "options" as external parameter that might read better might be found amongst "respecting" or "per" or "modifiedBy".

Sources/ExecutorType.swift
```diff
@@ -0,0 +1,68 @@
...
+//  ExecutorType.swift
+//  Deferred
+//
+//  Created by Zachary Waldowski on 3/29/16.
+//  Copyright © 2014-2016 Big Nerd Ranch. All rights reserved.
+//
+
+import Foundation
+
+public protocol ExecutorType {
```
jeremy-w: Public type - should have docs, no?

Sources/ExistentialFuture.swift
```diff
@@ -107,12 +105,13 @@ public struct Future<Value>: FutureType {
...
-    /// Call some function once the underlying future's value is determined.
+    /// Call some function `body` once the underlying future's value is
+    /// determined.
     ///
-    /// - parameter queue: A dispatch queue for executing the given function on.
-    /// - parameter function: A function that uses the determined value.
-    public func upon(queue: dispatch_queue_t, body: Value -> ()) {
-        return box.upon(queue, body: body)
+    /// If the value is already determined, the function will be submitted to
+    /// to `executor` immediately. An `upon` call is always asynchronous.
```
jeremy-w: > An upon call is always asynchronous.

:100: Happy to see this spelt out clearly!

Sources/FutureComposition.swift
```diff
@@ -0,0 +1,52 @@
+//
+//  FutureComposition.swift
+//  Deferred
+//
+//  Created by Zachary Waldowski on 4/2/16.
+//  Copyright © 2014-2016 Big Nerd Ranch. Licensed under MIT.
+//
+
+extension FutureType {
+    /// Composes this future with another.
```
jeremy-w: This description doesn't really tell me much - "composes" is too vague to be useful. The `returns` doc OTOH is great. I'd recommend making that the main comment, dropping the `parameter other` bit (it doesn't tell you anything the types don't already), and leaving the `seealso`.

Sources/FutureComposition.swift
```diff
@@ -0,0 +1,52 @@
...
+    ///
+    /// - parameter other: Any other future.
+    /// - returns: A value that becomes determined after both the reciever and
```
jeremy-w: typo: reciever -> receiver

Sources/FutureComposition.swift
```diff
@@ -0,0 +1,52 @@
...
+    /// - seealso: SequenceType.allFutures
+    public func and<OtherFuture: FutureType>(other: OtherFuture) -> Future<(Value, OtherFuture.Value)> {
+        return flatMap { t in other.map { u in (t, u) } }
+    }
+    
+    /// Composes this future with others.
+    ///
+    /// - parameter one: Some other future to join with.
+    /// - parameter two: Some other future to join with.
+    /// - returns: A value that becomes determined after the reciever and both
```
jeremy-w: same typo

Sources/FutureMap.swift
```diff
@@ -0,0 +1,45 @@
...
+//  FutureMap.swift
+//  Deferred
+//
+//  Created by Zachary Waldowski on 4/2/16.
+//  Copyright © 2014-2016 Big Nerd Ranch. Licensed under MIT.
+//
+
+extension FutureType {
+
+    public func map<NewValue>(upon executor: ExecutorType, _ transform: Value -> NewValue) -> Future<NewValue> {
```
jeremy-w: Undocumented public func - should be able to steal the docs for the queue-based version just below it in the file.

Sources/FutureType.swift
```diff
@@ -30,14 +30,14 @@ public protocol FutureType: CustomDebugStringConvertible, CustomReflectable {
     /// A type that represents the result of some asynchronous operation.
     associatedtype Value
 
-    /// Call some function once the value is determined.
+    /// Call some function `body` once the value is determined.
```
jeremy-w: s/function/closure/ ?

Sources/FutureType.swift
```diff
@@ -30,14 +30,14 @@ public protocol FutureType: CustomDebugStringConvertible, CustomReflectable {
...
     ///
-    /// If the value is determined, the function should be submitted to the
-    /// queue immediately. An `upon` call should always execute asynchronously.
+    /// If the value is determined, the function should be submitted to
+    /// to the `executor` immediately.
```
jeremy-w: This has half the verbiage of an earlier `upon`: this one specifies immediate submission, but not async execution. Could you clarify how the closure is executed?

Sources/Deferred.swift
zwaldowski: Hmm, very good point. Do you "notify **with** an executor", "notify an executor", "notify with an executor"?

Sources/Deferred.swift
zwaldowski: Yeah, `inOptions` is definitely not in line. Will think about `body`'s external parameter.

Sources/ExecutorType.swift
zwaldowski: I knew I forgot something.

Sources/FutureComposition.swift
zwaldowski: These are old typos! Aged, like a fine wine.

Sources/FutureType.swift
zwaldowski: This should come from documenting the intent of `ExecutorType`. Unfortunately, we can only be as strong about an `upon` always being asynchronous for the queue overloads; though I don't recommend it, you can totally write an `ExecutorType` that just calls its closure immediately.

Sources/FutureType.swift
zwaldowski: The stdlib docs seem to be coalescing around "the `body` closure" (following the new internal parameter rule). I'll see what I can do.

Sources/FutureComposition.swift
zwaldowski: Yeah, I think these can do away with the doc tags (except for `seealso`) and just be "Creates a future that becomes…"

[See also](https://swift.org/documentation/api-design-guidelines/):

> Describe what a function or method does and what it returns, omitting null effects and Void returns:
> 
> ```swift
> /// Inserts `newHead` at the beginning of `self`.
> mutating func prepend(newHead: Int)
> 
> /// Returns a `List` containing `head` followed by the elements
> /// of `self`.
> func prepending(head: Element) -> List
> 
> /// Removes and returns the first element of `self` if non-empty;
> /// returns `nil` otherwise.
> mutating func popFirst() -> Element?
> ```

zwaldowski:
Note to self (@zwaldowski): when doing the doc pass, remember to sweep for old-style AppleDoc comment tags.

Sources/Deferred.swift
zwaldowski: Not as concerned about these since they're internal API, but, yes, they need work.

zwaldowski:
Pushed documentation updates. Letting :construction_worker: have a look at some unit tests

Issue #84, opened 2016-03-27
[README] Table-of-contents-ify


zwaldowski:
As mentioned in [#78](https://github.com/bignerdranch/Deferred/pull/78#issuecomment-199319205):

> @jeremy-w: This README might have reached the point where a Table of Contents is appropriate. I generally use doctoc for that (`npm -g install tocdoc`).

[And](https://github.com/bignerdranch/Deferred/pull/78#issuecomment-199319694): 

> @zwaldowski: @jeremy-w To that point, the TOC'ed and usage-based readme should probably be transcluded in Jazzy, as if it were the "Deferred Type Programming Guide" (as long as we're going for Apple-like).

[And](https://github.com/bignerdranch/Deferred/pull/78#issuecomment-201189486):

> @zwaldowski: Totally outside the scope of this PR, but just for reference I had in mind what [ReSwift](http://reswift.github.io/ReSwift/master/) does — it's sort of the poster child for Jazzy. Notice that they split up their README into different Markdown documents that get used in the documentation set.

Issue #70, opened 2016-01-26
[README] Explain Future and Promise Types


randomstep:
Deferred preferred interface is through Future and Promise Types. We should update the README to point people toward this usage and appropriate documentation.

Short version: Future and Promise split the 2 sides of the Deferred API. Future is read-only, Promise is write-only. So you make a Deferred, ensure the Promise view has only one owner, and give the Future out to anyone that wants it.

zwaldowski:
In order to follow The Law of Conservation of Existing Sections of the README, doing an additive approach until we can come back and weave `Future` into the narrative more.

PR #94, opened 2016-04-11
Documentation and README gardening

zwaldowski:
Hi there, 2.0! (More or less)

Address #70 and #84 through improvements minor changes to documentation, improvements to doc-set generation, and README reworking.

#### Testing

Not applicable.

#### API Changes

Not applicable.

### DIFF
--- a/.jazzy.yml
+++ b/.jazzy.yml
@@ -2,6 +2,35 @@
 author: "Big Nerd Ranch"
 author_url: "http://bignerdranch.github.io"
 copyright: "© 2014-2016 [Big Nerd Ranch](https://www.bignerdranch.com) under [MIT license](https://github.com/bignerdranch/Deferred/blob/master/LICENSE.txt)."
+custom_categories:
+  - name: Futures and Promises
+    children:
+    - Deferred
+    - FutureType
+    - PromiseType
+    - Timeout
+  - name: Future Mastery
+    children:
+    - Future
+    - CollectionType
+    - SequenceType
+    - IgnoringFuture
+  - name: Custom Execution
+    children:
+    - ExecutorType
+    - NSOperationQueue
+    - NSRunLoop
+    - CFRunLoop
+  - name: Data Locking
+    children:
+    - LockProtected
+    - ReadWriteLock
+  - name: Locks
+    children:
+    - DispatchLock
+    - SpinLock
+    - CASSpinLock
+    - PThreadReadWriteLock
 github_url: "https://github.com/bignerdranch/Deferred"
 head: |
   <meta property="og:site_name" content="Deferred programming Guide">
--- a/.jazzy.yml
+++ b/.jazzy.yml
@@ -1,0 +1,21 @@
+# jazzy --swift-version 2.2 -m Deferred -g "https://github.com/bignerdranch/Deferred" -a "Big Nerd Ranch" -u "http://bignerdranch.github.io" --module-version=2.0.0 -r "http://bignerdranch.github.io/Deferred/"
+author: "Big Nerd Ranch"
+author_url: "http://bignerdranch.github.io"
+copyright: "© 2014-2016 [Big Nerd Ranch](https://www.bignerdranch.com) under [MIT license](https://github.com/bignerdranch/Deferred/blob/master/LICENSE.txt)."
+github_url: "https://github.com/bignerdranch/Deferred"
+head: |
+  <meta property="og:site_name" content="Deferred programming Guide">
+  <meta property="og:description" content="An implementation of OCaml's Deferred for Swift. It lets you work with values that haven't been determined yet, like an array that's coming later from a web service call.">
+  <meta property="og:image" content="https://www.bignerdranch.com/img/bnr-logo-square.png">
+  <link rel="shortcut icon" type="image/x-icon" href="https://www.bignerdranch.com/favicon.png">
+hide_documentation_coverage: true
+module: Deferred
+module_version: "2.0.0"
+# podspec: BNRDeferred.podspec
+root_url: "http://bignerdranch.github.io/Deferred/"
+swift_version: 2.2
+theme: fullwidth
+
+# CocoaDocs currently using an old Jazzy
+module_name: Deferred
+author_name: Big Nerd Ranch
--- a/.jazzy.yml
+++ b/.jazzy.yml
@@ -1,4 +1,4 @@
-# jazzy --swift-version 2.2 -m Deferred -g "https://github.com/bignerdranch/Deferred" -a "Big Nerd Ranch" -u "http://bignerdranch.github.io" --module-version=2.0.0 -r "http://bignerdranch.github.io/Deferred/"
+# jazzy --config .jazzy.yml
 author: "Big Nerd Ranch"
 author_url: "http://bignerdranch.github.io"
 copyright: "© 2014-2016 [Big Nerd Ranch](https://www.bignerdranch.com) under [MIT license](https://github.com/bignerdranch/Deferred/blob/master/LICENSE.txt)."
--- a/.jazzy.yml
+++ b/.jazzy.yml
@@ -1,7 +1,7 @@
 # jazzy --config .jazzy.yml
 author: "Big Nerd Ranch"
 author_url: "http://bignerdranch.github.io"
-copyright: "© 2014-2016 [Big Nerd Ranch](https://www.bignerdranch.com) under [MIT license](https://github.com/bignerdranch/Deferred/blob/master/LICENSE.txt)."
+copyright: "© 2014-2016 [Big Nerd Ranch](//www.bignerdranch.com) under [MIT license](//github.com/bignerdranch/Deferred/blob/master/LICENSE.txt)."
 custom_categories:
   - name: Futures and Promises
     children:
@@ -35,8 +35,8 @@ github_url: "https://github.com/bignerdranch/Deferred"
 head: |
   <meta property="og:site_name" content="Deferred Programming Guide">
   <meta property="og:description" content="An implementation of OCaml's Deferred for Swift. It lets you work with values that haven't been determined yet, like an array that's coming later from a web service call.">
-  <meta property="og:image" content="https://www.bignerdranch.com/img/bnr-logo-square.png">
-  <link rel="shortcut icon" type="image/x-icon" href="https://www.bignerdranch.com/favicon.png">
+  <meta property="og:image" content="//www.bignerdranch.com/img/bnr-logo-square.png">
+  <link rel="shortcut icon" type="image/x-icon" href="//www.bignerdranch.com/favicon.png">
 hide_documentation_coverage: true
 module: Deferred
 module_version: "2.0.0"
--- a/.jazzy.yml
+++ b/.jazzy.yml
@@ -33,7 +33,7 @@ custom_categories:
     - PThreadReadWriteLock
 github_url: "https://github.com/bignerdranch/Deferred"
 head: |
-  <meta property="og:site_name" content="Deferred programming Guide">
+  <meta property="og:site_name" content="Deferred Programming Guide">
   <meta property="og:description" content="An implementation of OCaml's Deferred for Swift. It lets you work with values that haven't been determined yet, like an array that's coming later from a web service call.">
   <meta property="og:image" content="https://www.bignerdranch.com/img/bnr-logo-square.png">
   <link rel="shortcut icon" type="image/x-icon" href="https://www.bignerdranch.com/favicon.png">
--- a/.jazzy.yml
+++ b/.jazzy.yml
@@ -9,23 +9,23 @@ custom_categories:
     - FutureType
     - PromiseType
     - Timeout
-  - name: Future Mastery
+  - name: Transforming Future
     children:
     - Future
     - CollectionType
     - SequenceType
     - IgnoringFuture
-  - name: Custom Execution
+  - name: Controlling Where Callbacks Execute
     children:
     - ExecutorType
     - NSOperationQueue
     - NSRunLoop
     - CFRunLoop
-  - name: Data Locking
+  - name: Coordinating Access to Data
     children:
     - LockProtected
     - ReadWriteLock
-  - name: Locks
+  - name: Coordination Primitives
     children:
     - DispatchLock
     - SpinLock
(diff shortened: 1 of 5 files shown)


zwaldowski:
Be sure to check out the [rich diff](https://github.com/bignerdranch/Deferred/pull/94/files?short_path=04c6e90#diff-04c6e90faac2675aa89e2176d2eec7d8) view of the README! :smile:

davidahouse:
@zwaldowski Looks great!

In the Combining Deferreds section, should this line be `joinedValues` instead of `all`? Or should we change the example to use `all` instead? Or should we have an example of both?

```
// `all` creates a new Deferred that is filled once all inputs are available.

...

let allDeferreds: Future<[Int]> = deferreds.joinedValues
```

zwaldowski:
`joinedValues` replaces `all`. Good catch!

davidahouse:
Found another cleanup item. The `Package.swift` file contains the wrong url. It should be:

```
.Package(url: "https://github.com/bignerdranch/Deferred.git", majorVersion: 1)
```

zwaldowski:
Nope, this upcoming release is 2.0.

davidahouse:
@zwaldowski OK, so it has TWO problems. :)

currently is:
```
 .Package(url: "https://github.com/bignerdranch/AtomicSwift.git", majorVersion: 1)
```

zwaldowski:
**Oh** you were looking at the Package.swift.

No, the SPM manifest is correct. Deferred itself is a package that depends (for now) on the [AtomicSwift](https://github.com/bignerdranch/AtomicSwift/blob/master/Package.swift) pseudo-package.

davidahouse:
@zwaldowski Ahh right, in the dependencies section. Sorry was scanning and missed the context. Carry on!

.jazzy.yml
```diff
@@ -0,0 +1,50 @@
+# jazzy --swift-version 2.2 -m Deferred -g "https://github.com/bignerdranch/Deferred" -a "Big Nerd Ranch" -u "http://bignerdranch.github.io" --module-version=2.0.0 -r "http://bignerdranch.github.io/Deferred/"
```
jeremy-w: For scripts, I prefer to use the long version of options where possible. The abbreviation when typing them out yourself of short options isn't needed, and the explanatory power of long options is wonderful.

If jazzy'll let you, I often spread args in a group across lines using continuation backslashes. (This happens a lot with gnarly `curl` commands.) Here's an example of what I mean from a Makefile I have lying around:

```make
run:
	browser-sync start \
		--server \
		--files "stylesheets/*.css, scripts/*.js, *.html"
```

jeremy-w: On second thought, looks like this comment line was for copy-paste execution at the CLI. But it's out-of-date now we need so many options passed to jazzy. Nuke the line, or explain how to invoke it with this config file?

zwaldowski: Ah! I didn't mean to leave this in. It should be sufficient to use the config file. Will note that instead.

.jazzy.yml
```diff
@@ -0,0 +1,50 @@
...
+author: "Big Nerd Ranch"
+author_url: "http://bignerdranch.github.io"
+copyright: "© 2014-2016 [Big Nerd Ranch](https://www.bignerdranch.com) under [MIT license](https://github.com/bignerdranch/Deferred/blob/master/LICENSE.txt)."
```
jeremy-w: I'd make this "(C) DATE COMPANY. Licensed under the terms of the [MIT license]."

I also tend to use scheme-relative links like `//www.bignerdranch.com`, but that's not a big deal.

I'd be tempted to link the MIT license bit to the TLDR;Legal summary page: https://tldrlegal.com/license/mit-license#summary

zwaldowski: Are scheme-relative links valid Markdown?

zwaldowski: Doesn't seem like it http://meta.stackexchange.com/questions/121293/markdown-support-for-urls-without-a-protocol

zwaldowski: Sorry, right. Clarification. The `copyright` stanza is included as Markdown in the footer of the docs.

.jazzy.yml
```diff
@@ -0,0 +1,50 @@
...
+custom_categories:
+  - name: Futures and Promises
+    children:
+    - Deferred
+    - FutureType
+    - PromiseType
+    - Timeout
+  - name: Future Mastery
```
jeremy-w: This category stands out oddly amongst the other categories. I'm not sure what to rename it, or how Future contrasts to FutureType.

zwaldowski: Yeah, I'm unsure as well. "Things That Are Futures But Are Not Deferred" seemed too wordy.

.jazzy.yml
```diff
@@ -0,0 +1,50 @@
...
+    - CollectionType
+    - SequenceType
+    - IgnoringFuture
+  - name: Custom Execution
+    children:
+    - ExecutorType
+    - NSOperationQueue
+    - NSRunLoop
+    - CFRunLoop
+  - name: Data Locking
```
jeremy-w: Suggest "Coordinating Access to Data".

.jazzy.yml
```diff
@@ -0,0 +1,50 @@
...
+    children:
+    - LockProtected
+    - ReadWriteLock
+  - name: Locks
```
jeremy-w: Suggest "Coordination Primitives".

.jazzy.yml
zwaldowski: :+1:

.jazzy.yml
zwaldowski: :OK:

.jazzy.yml
```diff
@@ -0,0 +1,50 @@
...
+    - FutureType
+    - PromiseType
+    - Timeout
+  - name: Future Mastery
+    children:
+    - Future
+    - CollectionType
+    - SequenceType
+    - IgnoringFuture
+  - name: Custom Execution
```
jeremy-w: Suggest "Controlling Where Callbacks Execute".

jeremy-w:
Is there somewhere I can view the prepped docs online, so I can see how the categories look in context?

zwaldowski:
I'll push it up to a bucket really quick

README.md
```diff
@@ -1,82 +1,67 @@
...
-```
-
-You will also need to make sure you're opting into using frameworks:
-
-```ruby
-use_frameworks!
-```
-
-Then run `pod install`.
+This is an implementation of [OCaml's Deferred](https://ocaml.janestreet.com/ocaml-core/111.25.00/doc/async_kernel/#Deferred) for Swift. It lets you work with values that haven't been determined yet, like an array that's coming later (one day!) from a web service call.
```
jeremy-w: Let's reverse these two sentences:

> Deferred lets you work…call. It was originally inspired by [OCaml's Deferred library.](https://ocaml.janestreet.com/ocaml-core/111.25.00/doc/async_kernel/#Deferred)

README.md
```diff
@@ -1,82 +1,67 @@
...
+
+### Vital Statistics
+
+|             | **Status**                                                                                                                                                  |
+|-------------|-------------------------------------------------------------------------------------------------------------------------------------------------------------|
```
jeremy-w: This table is marvelously unreadable in monospace. :)

README.md
```diff
@@ -131,23 +115,23 @@ This gets really messy with callbacks. But with `Deferred`, it's just:
...
     // names has been determined - use it!
     dataSource.array = names
     tableView.reloadData()
 }
 ```
 
-## Tasks
+## Basic Tasks
 
 ### Vending a Future
```
jeremy-w: Now that we have an actual `Future` type, we should probably be more careful about Future/Deferred verbiage. We probably do want to counsel them to vend a Future and not a Deferred, though.

README.md
```diff
@@ -322,9 +321,154 @@ func AsynchronousFriends(forUser: jimbob) -> Deferred<[Friend]> {
...
-### Further Information
+## Mastering (the) `Future`
+
+Deferred is designed to scale with the fundamentals you see above. Large applications can be built using just `Deferred` and its `upon` and `fill` methods.
+
+### Read-Only Views
+
+It sometimes just doesn't make *sense* to be able to `fill` something; if you have a `Deferred` wrapping `UIApplication`'s push notification token, what does it mean if someone in your codebase calls `fill` on it?
+
+You may have noticed that anybody can call `upon` and `fill` on `Deferred` at any time; it's **read-write**. The former is fundamental, but the latter may be a liability as different pieces of code interact with each other. There's no concern to be had about thread safety, but instead about access: how do you limit illogical control flow, make it **read-only**?
```
jeremy-w: It's not clear to me what "former" and "latter" refer to here. Rewrite to avoid them.

README.md
zwaldowski: Make the width _really_ big. :grimacing:

(Maybe I can use a URL shortener.)

README.md
jeremy-w: I don't get the "illogical control flow" bit. What control flow? I think that's unnecessary though. We can replace this whole sentence with just, "How can we make it read-only?"

zwaldowski:
https://bucket.waldowski.me/DeferredDocs20160412

README.md
zwaldowski: I think we would be well served by

* "Future" (the idea)
* "the `Future` type" or "the `Future<Value>` type"
* "Deferred" (the framework)
* "the `Deferred` type" or "the `Deferred<Value>` type"

.jazzy.yml
jeremy-w: It does. The answer you wanted to read is:
http://meta.stackexchange.com/a/121321 The asker wasn't writing a valid
URL; the scheme-relative ones begin with a leading //, not directly with
the hostname.

I'm using it already on several GitHub READMEs, so. :) To see it in action,
check the raw view here:
https://github.com/bignerdranch/jws-coffeerun/blob/master/README.md

--
Jeremy W. Sherman <jeremy@bignerdranch.com>

2016-04-12 13:41 GMT-04:00 Zachary Waldowski <notifications@github.com>:

zwaldowski: OK, sweet. I'll change it

README.md
jeremy-w: I think using the footnote-style version would work, like:

[mumble][1]

  [1]: url goes here

That can be a string rather than a 1, and I recommend that in order to
avoid conflicts. :) You can stick the link-footnotes right under the table
so they stay by the text using them.

--
Jeremy W. Sherman <jeremy@bignerdranch.com>

2016-04-12 13:50 GMT-04:00 Zachary Waldowski <notifications@github.com>:

README.md
jeremy-w: I like that. Maybe even go with "Vending a Future Value" to stick with the
"this is a value that hasn't quite arrived yet". Then we get future value
vs future type verbal distinction, even absent any fancy markup.

--
Jeremy W. Sherman <jeremy@bignerdranch.com>

2016-04-12 13:58 GMT-04:00 Zachary Waldowski <notifications@github.com>:

README.md
zwaldowski: I always forget about those! Will play with it, thanks

.jazzy.yml
jeremy-w: Maybe "Controlling Who Can Fill" or "Access Control"? And shift `FutureType` in here alongside `Future`?

zwaldowski: I'm not sure that approach is right for the framework at large, it just made most sense for the README.

zwaldowski: Eh. Maybe you're right.

.jazzy.yml
zwaldowski: This conveniently breaks when I'm viewing locally (`file://www.bignerdranch.com`), but so be it

jeremy-w: Ah, fun. Try viewing on `http://localhost:PORT` via `python -mSimpleHTTPServer` or [devd](https://github.com/cortesi/devd) perhaps?

README.md
```diff
@@ -329,7 +329,7 @@ Deferred is designed to scale with the fundamentals you see above. Large applica
 
 It sometimes just doesn't make *sense* to be able to `fill` something; if you have a `Deferred` wrapping `UIApplication`'s push notification token, what does it mean if someone in your codebase calls `fill` on it?
 
-You may have noticed that anybody can call `upon` and `fill` on `Deferred` at any time; it's **read-write**. The former is fundamental, but the latter may be a liability as different pieces of code interact with each other. There's no concern to be had about thread safety, but instead about access: how do you limit illogical control flow, make it **read-only**?
+You may have noticed that anybody can call `upon` on a `Deferred` type; this is fundamental. But the same is true of `fill`, and this may be a liability as different pieces of code interact with each other. How can we make it **read-only**?
```
jeremy-w: This is much clearer. :grapes:

Sources/ExecutorType.swift
```diff
@@ -55,6 +83,12 @@ struct QueueExecutor: ExecutorType {
 
 }
 
+/// An operation queue manages a number of operation objects, making higher-
```
jeremy-w: Does this correctly hyphenate as `higher-level` in Xcode and Jazzy, or does it end up rendering as `higher- level` with a space in the middle due to the linebreak?

README.md
```diff
@@ -322,9 +326,154 @@ func AsynchronousFriends(forUser: jimbob) -> Deferred<[Friend]> {
...
+
+It sometimes just doesn't make *sense* to be able to `fill` something; if you have a `Deferred` wrapping `UIApplication`'s push notification token, what does it mean if someone in your codebase calls `fill` on it?
+
+You may have noticed that anybody can call `upon` on a `Deferred` type; this is fundamental. But the same is true of `fill`, and this may be a liability as different pieces of code interact with each other. How can we make it **read-only**?
+
+For this reason, Deferred is split into `FutureType` and `PromiseType`, both protocols the `Deferred` type conforms to. You can think of these as the "reading" and "writing" sides of a deferred value; a future can only be `upon`ed, and a promise can only be `fill`ed.
+
+Deferred also provides the `Future` type, a wrapper for anything that's a `FutureType` much like the Swift standard library's `Any` types. You can use it protectively to make a `Deferred` read-only. Reconsider the example from above:
+
+```
```
jeremy-w: Should we ensure all code blocks are have the language marked, as:

    ```swift

and not just:

    ```

?

Sources/ExecutorType.swift
zwaldowski: Hilariously, neither. The "NSOperationQueue Extension Reference" does not include this doc at all.

Moved, it is the dumb `higher- level`. I'll fix it.

README.md
zwaldowski: Fenced!

PR #98, opened 2016-04-18
Task

zwaldowski:

See #64.

Introduces a `TaskType`, defining a cancellable `Future` over a `Result`.

`Task` (mirroring `Future`) is the main entry point into this API.

#### Testing

Cleans up some shared fixtures between Deferred and Task unit tests.

- [x] Improve test coverage of `Task` and its various and sundry extensions.
- [x] Improve test coverage of `FutureType where Value: ResultType` extensions.

The new additions have reasonably high test coverage.

#### API Changes

Changes are entirely additive to Deferred 2.0.


Sources/Result/ResultType.swift
```diff
@@ -0,0 +1,41 @@
...
+//  Deferred
+//
+//  Created by Zachary Waldowski on 12/9/15.
+//  Copyright © 2014-2016 Big Nerd Ranch. Licensed under MIT.
+//
+
+/// A type that can exclusively represent either some result value of a
+/// successful computation or a failure with an error.
+public protocol ResultType: CustomStringConvertible, CustomDebugStringConvertible {
+    associatedtype Value
```
zwaldowski: I invite improvements to the name of this associated type. Bike shedders, activate!

Sources/Task/ExistentialTask.swift
```diff
@@ -0,0 +1,93 @@
...
+        cancellation()
+    }
+}
+
+extension Task {
+    /// Create a task whose `upon(_:body:)` method uses the result of `base`.
+    public init<Task: FutureType where Task.Value: ResultType, Task.Value.Value == SuccessValue>(_ base: Task, cancellation: Cancellation) {
+        self.init(base.map {
+            Value(with: $0.extract)
+        }, cancellation: cancellation)
```
zwaldowski: This could/should be a `lazy` map if/when we resurrect #75.

Sources/Task/ExistentialTask.swift
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    /// Create a task whose `upon(_:body:)` method uses the result of `base`.
+    public init<Task: TaskType where Task.Value.Value == SuccessValue>(_ base: Task) {
+        self.init(base.map {
+            Value(with: $0.extract)
+        }, cancellation: base.cancel)
```
zwaldowski: Same above as with `lazy`.

Sources/Task/ExistentialTask.swift
zwaldowski: Strike that, these should use the extension we define on `ResultType` in `ResultFuture.swift`.

Sources/Task/ResultFuture.swift
```diff
@@ -0,0 +1,85 @@
...
+        upon(Self.genericQueue, body: commonUponFailure(body))
+    }
+}
+
+extension Future where Value: ResultType {
+    /// Create a future having the same underlying task as `other`.
+    public init<Other: FutureType where Other.Value: ResultType, Other.Value.Value == Value.Value>(_ other: Other) {
+        self = other.map {
+            Value(with: $0.extract)
+        }
```
zwaldowski: This could/should be a `lazy` map if/when we resurrect #75.

Sources/Task/ResultPromise.swift
```diff
@@ -0,0 +1,36 @@
...
+import Result
+#endif
+
+extension PromiseType where Value: ResultType {
+    /// Completes the task with a successful `value`, or a thrown error.
+    ///
+    /// - seealso: `fill(_:)`
+    public func succeed(@autoclosure value: () throws -> Value.Value) -> Bool {
+        return fill(Value(value: value))
+    }
```
zwaldowski: Like the `ResultType` initializer, I don't think this works like we want it to. Need to test if `future.succeed(try foo())` (or `try future.succeed(foo())`?) still requires the surrounding context be `throws`.

zwaldowski: If it doesn't work, that's probably a compiler bug.

davidahouse:
I wrote some introductory tests for Task today and so far this work looks good. One question is where you plan to go with the `TaskAccumulator` and `TaskCollections`? These feel like the same kind of thing in general with very different implementations (and of course the difference between if it cares about the result or not). Overall I like the thought of using `TaskCollections` as a way to get a single Task that embodies some combination of an array of tasks (all finished, all success, first finished, etc). But you might have a more grand vision planed. Please share. :)

zwaldowski:
There's not yet an over-arching vision for it, just that they were both in our original distribution of `Task`. We have definitely considered/weighed the different paths you'd want. The distinctions are fairly subtle.

Currently, there is:
* With `[Task<T>]`, use `joinedValues` to know when everything _finishes_, getting a `Future<[Result<T>]>`.
   - We could perhaps shadow this to make the result a cancellable `Task<[T]>`.
* With `[Task<T>]`, use `joinedTasks` to know when everything _succeeds_, getting a `Task<Void>`.
* With an incremental list of any `Task` (for example, a Core Data store that vends `Task`s over time), use `TaskAccumulator` to know when everything currently in-flight has _finished_.

@jgallagher can speak better to the idea of `TaskAccumulator`, and weigh in on whether we need it.

davidahouse:
Cool sounds good! I'll try that joinedValues. That is one thing I was looking for and didn't find.

Sources/Task/ResultPromise.swift
zwaldowski: Seems fine to me. You can't `throw` inside the parens, though, which is to be expected.

Sources/Result/ResultType.swift
zwaldowski: On hold until Swift 3

zwaldowski:
I'm splitting `NSURLSession` bits into #109 to maybe get this train moving; it was the piece lacking testing entirely.

Sources/Task/ResultPromise.swift
zwaldowski: I've identified other places `@autoclosure` breaks for this, like the `ResultType` initializer. Swift 3 it is…

zwaldowski:
Coverage is up to 75%. Covering `TaskCollections.swift` and `IgnoringTask.swift` are my last big-ticket items, if anyone wanted to hit those before I can.

Tasking off WIP; ready for review.

Configurations/Base.xcconfig
```diff
@@ -39,8 +41,6 @@ TVOS_DEPLOYMENT_TARGET = 9.0
 ENABLE_BITCODE = YES
 ENABLE_BITCODE[sdk=macosx*] = NO
 
-CODE_SIGN_IDENTITY[sdk=macosx*] = -
-CODE_SIGN_IDENTITY[sdk=iphoneos*] = iPhone Developer
-CODE_SIGN_IDENTITY[sdk=appletvos*] = iPhone Developer
+CODE_SIGN_IDENTITY =
```
zwaldowski: Drop this, only for Xcode 8/Swift 3

PR #99, opened 2016-04-29
Remove overridden mirror for Future<Value>

zwaldowski:

Removes an override introduced in #80.

Before the PR, `Future<T>` would appear in a playground as `Deferred.BlahFutureBox<T>SomeUUID`. That's not good. Now it returns to the default `FutureType` implementation, showing `(isFilled **false**)` etc.

#### Testing

As noted in #80, mirrors are not to be tested.

#### API Changes

Only affects debugging.


jeremy-w:
Nothing to argue about here. :D

Issue #101, opened 2016-07-07
Apply Swift 3 API Guidelines


zwaldowski:
Bikeshed time. See also #20, #21, and #66.

Are you a BNR person assigned to this issue? You've been assigned for review! You don't need to do anything other than read through this issue and leave feedback.

This document refers to the Swift 3-style name of types unless explicitly referring to the old API.

### Proposed changes

<details>
* `Deferred.init(value:)` ➡️ `Deferred.init(filledWith:)`
* `ExecutorType` ➡️ `Executor`
* `ExecutorType.submit(_:)` ➡️ `Executor.async(_:)`
* `Sequence<FutureType>.earliestFilled` ➡️ `Sequence<FutureProtocol>.firstCompleted()`
* `Collection<FutureType>.joinedValues` ➡️ `Collection<FutureProtocol>.allCompleted()`
* `FutureType` ➡️ `FutureProtocol`
* `FutureType.wait(_:)` ➡️ `FutureProtocol.wait(until:)`
* `FutureType.Type.genericQueue` ➡️ `DispatchQueue.Type.any()` (i.e., `upon(.any()) { ... }`)
* `FutureType.uponMainQueue(_:)` ➡️ removed (use `upon(.main)`)
* `FutureType.upon(:_:)` ➡️ `FutureType.upon(_:execute:)`
* `FutureType.map(upon:_:)` ➡️ `FutureType.map(upon:transform:)`
* `FutureType.flatMap(upon:_:)` ➡️ `FutureType.andThen(upon:start:)`
* `IgnoringFuture` ➡️ remove (#75)
* `LockProtected<T>` ➡️ `Protected<Value>`
* `PromiseType` ➡️ `PromiseProtocol`
* `PromiseType.fill(_:)` ➡️ `PromiseProtocol.fill(with:)`
* `PromiseType.fill(_:assertIfFilled:file:line:)` ➡️ `PromiseProtocol.mustFill(with:)`
* `ReadWriteLock` ➡️ `Locking`
* `Timeout` ➡️ removed, use Dispatch timeouts

#### Task
* introducing ➡️ `Task.init(executing:upon:)`
* `Task.init(upon:per:onCancel:body:)` ➡️ `Task.init(executedUpon:per:cancellingWith:body:)`
* `IgnoringTask` ➡️ removed (see also #75)
* `PromiseType<ResultType>.succeed(_:)` ➡️ `PromiseProtocol<ResultProtocol>.succeed(with:)`
[...]

PR #109, opened 2016-08-14
Task for NSURLSession

zwaldowski:
See #98.

Initially intended for inclusion in #98 (and, indeed, the genesis of the idea of `Task` entirely), this PR needs some separate consideration.

It needs another pass at design; I'm particularly concerned at losing the original `NSURLSessionTask` reference, the lack of progress reporting, and the lack of compatibility with delegate-based URL sessions.

#### Testing

- [ ] Needs a major pass at test coverage.

#### API Changes

Additions are additive.

API will naturally be broken majorly for Swift 3. See #101.


zwaldowski:
It was discussed today that this implementation falls down when using a session delegate (increasingly common for many features), and would thus be too coupled to encourage use.

We might consider an alternate design (such as a "`Task` tracker for your `NSURLSessionDelegate` implementation); we might consider skipping this MR and leaving its implementation in a FTMC on the wiki, or whatever.

zwaldowski:
I'm leaning towards the latter approach. Currently working on integrating this into [a wiki page](https://github.com/bignerdranch/Deferred/wiki/Working-with-URLSession). Thanks for the idea, @jjmanton.

PR #111, opened 2016-08-24
Task: Progress Reporting

zwaldowski:

Resurrecting #110. Reworks Task (see #98) to have progress reporting.

The primary semantics change that I can figure is that `cancel` handlers are executed asynchronously (all cancellation for a given progress tree is issued synchronously, though), which is in line with our "best attempt" recommendation.

Since there *are* semantics changes, I would like it to be reviewed separately.

#### Testing

See semantics changes note above.

Testing for this also uncovered some racing that occurred on Travis (or similarly constrained cases), and has been addressed.

#### API Changes

Though internal reworking has been done, the changes are mostly additive. The latest additions to the branch remove `TaskType`, which I can restore with an `@available` tag if deemed necessary.


Sources/Task/ExistentialTask.swift
```diff
@@ -10,29 +10,35 @@
...
-    private let cancellation: Cancellation
+    public let progress: NSProgress
 
-    /// Creates a task given a `future` and an optional `cancellation`.
-    public init(_ future: Future<Result>, cancellation: Cancellation) {
+    /// Creates a task given a `future` and its `progress`.
+    public init(future: Future<Result>, progress: NSProgress) {
         self.future = future
-        self.cancellation = cancellation
+        self.progress = NSProgress.extendingRoot(for: progress)
```
zwaldowski: Maybe `extendingRoot` should move into the `map`, `flatMap`, or `recover`. This design is flawed for allowing custom implementers of `TaskType`.

zwaldowski: Addressed!

Sources/Task/ExistentialTask.swift
```diff
@@ -10,29 +10,34 @@
...
-/// A simple function type expressing the cancellation of some operation.
-public typealias Cancellation = () -> Void
+import Foundation
 
 /// A wrapper over any task.
 ///
 /// Forwards operations to an arbitrary underlying future having the same result
 /// type, optionally combined with some `cancellation`.
-public struct Task<SuccessValue> {
+public final class Task<SuccessValue>: NSObject, NSProgressReporting {
```
zwaldowski: Maybe I should start leaving around `XPLAT` TODOs around for 3.x. This should stay a `struct` on Linux.

jeremy-w:
I reviewed the tests. If there are changes in semantics, the tests must have failed to capture the prior behavior, and so continue to fail to capture the current.

Can we write a test specifying the behavior so that we can ensure we don't accidentally break it in future?

zwaldowski:
@jeremy-w The tests capture the behavior of `cancel` perfectly fine. Derived `cancel`s were called asynchronously in the past, and they're called asynchronously now.

jeremy-w:
I'm confused, then: Your PR description says there's a semantics change. No test changed except by some method renaming and timeout extensions. …?

zwaldowski:
There is a *minor* semantics change. Before, calling `cancel` would cancel the underlying chain of Tasks at some indeterminate point in the future. Now, calling `cancel` will submit all cancellation handlers at once to a dispatch queue, meaning they will be called at some indeterminate point in the future. Any test that would've been written before that wouldn't work now would've been relying on implementation details.

zwaldowski:
In effect, we're making a guarantee that we weren't before, thanks to Foundation:

> Your block won't be invoked on any particular queue. If it must do work on a specific queue then it should schedule that work on that queue.

and its implementation (pseudo-code):

```swift
func cancel() {
    self.safeLockedSetValue(true, forKey: "isCancelled")

    lock.lock()
    defer { lock.unlock() }

    // We make no guarantees about on which queue handers are invoked.
    if let cancellationHandler = cancellationHandler {
        DispatchQueue.genericMatchingCurrent.async(execute: cancellationHandler)
    }
    
    for child in children {
        child.cancel()
    }
}
```

jeremy-w:
Ah, so what is missing is tests covering the interaction between task and progress reporting, then?

zwaldowski:
That's right. The branch was originally intended as a PoC and I was trying to validate some technique, but it's been working really well in playing around. I've added some (including fixing some behavior, as per usual 😅). Working on adding more!

Tests/TaskTests.swift
```diff
@@ -165,6 +165,89 @@ class TaskTests: XCTestCase {
...
+    }
+
+    func testThatTaskWrappingFilledIsDeterminate() {
+        let d = Deferred<Task<Int>.Result>.init(value: .Success(42))
+        let task = Task(d, cancellation: {})
+
+        XCTAssertFalse(task.progress.indeterminate)
+    }
+
+    func testThatMapExtendsParentFractions() {
```
jeremy-w: It's not clear to me what it means to extend a fraction. The code didn't clear it up for me.

Tests/TaskTests.swift
```diff
@@ -165,6 +165,89 @@ class TaskTests: XCTestCase {
...
+        let beforeTask: Task<Int> = Task(value: 42)
+        XCTAssertEqualWithAccuracy(beforeTask.progress.fractionCompleted, 1, accuracy: 0.001)
```
jeremy-w: This assertion looks like it belongs in a separate `testTaskCreatedFilledIs100PercentCompleted()`.

`1.00` should also be representable without loss of accuracy by both Float and Double, so we shouldn't need any `accuracy` wiggle-room here.

Tests/TaskTests.swift
```diff
@@ -165,6 +165,89 @@ class TaskTests: XCTestCase {
...
+
+        waitForExpectationsWithTimeout(TestTimeout, handler: nil)
+    }
+
+    func testThatFlatMapExtendsParentFractions() {
+        let beforeTask: Task<Int> = Task(value: 42)
+        XCTAssertEqualWithAccuracy(beforeTask.progress.fractionCompleted, 1, accuracy: 0.001)
+
+        let afterTask = beforeTask.flatMap { (before) -> Task<Int> in
+            let d = Deferred<Task<Int>.Result>()
```
jeremy-w: There's a lot going on in this `flatMap`, and it's not clear to me what the intention is, only the implementation. It's also hard to tell what's essential detail vs incidental detail (i.e., "we had to pick some value"). Sometimes creating helper functions named for the intention can make the test more readily comprehensible; that might help here.

Tests/TaskTests.swift
```diff
@@ -165,6 +165,89 @@ class TaskTests: XCTestCase {
...
+
+    func testThatTaskWrappingFilledIsDeterminate() {
+        let d = Deferred<Task<Int>.Result>.init(value: .Success(42))
+        let task = Task(d, cancellation: {})
+
+        XCTAssertFalse(task.progress.indeterminate)
+    }
+
+    func testThatMapExtendsParentFractions() {
+        let beforeTask: Task<Int> = Task(value: 42)
```
jeremy-w: Make have a `var anyFilledTask: Task<Int> { return Task(value: 42) }` on the test case class?

jeremy-w: (For that matter, would we say "filled task" or "finished task"?)

Tests/TaskTests.swift
zwaldowski: That's fair. Will add more prose around this throughout the codebase.

30 sec summary (as I'm on my phone) is that NSProgress is a tree, and Task mapping is linear. Each new transform adds 1 unit to the denominator of the "root" Task.

Tests/TaskTests.swift
zwaldowski: Great points

Tests/TaskTests.swift
zwaldowski: Yeah, I meant to factor that out. Most of these are similarly contrived.

Tests/TaskTests.swift
zwaldowski: I'd say completed Task, fulfilled Future, filled Promise.

PR #112, opened 2016-09-11
Deferred 3.0

zwaldowski:

Deferred 3 is a major API break with Deferred 2, adopting the [API Design Guidelines](https://swift.org/documentation/api-design-guidelines/) and naming conventions, using the latest and greatest of Grand Central Dispatch, and deploying on Linux for the first time.
#### Testing

At the time of writing, testing coverage will remain largely the same as in 2.x. The removal of some overloads has allowed removing some test cases entirely without a change of coverage.

Travis and the Xcode 8 GM are currently having issues running the tests against iOS 8, thus that testing target is dropped for now.
#### API Changes

Everything is broken. Deferred 3.0 will be a major API break, although mostly in terms of simple naming.


piercifani:
This looks awesome! Any ETA on this?

zwaldowski:
@piercifani I'll cut a beta once we review and merge #119.

zwaldowski:
Thanks @rcedwards! I'll `bundle update` and see if we can escape that.

Issue #137, opened 2016-10-27
[3.0.0beta1] Crash on TaskProgress


piercifani:
I cannot recreate it on a smaller project. But I'm getting consistent crashes on latest commit of the Swift 3 branch. The backtrace looks like this: ![](https://i.imgur.com/ZVLfdw0.jpg)

The logger says:

```_TtC8DeferredP33_9528CFABF28443747CE301F7DAD432BB13ProxyProgress object 0x127c17fa0 overreleased while already deallocating; break on objc_overrelease_during_dealloc_error to debug```

Building for ARM64 in DEBUG. Please let me know if there is a way to help debug this, and as always: THANKS!

jeremy-w:
Please set a breakpoint on `objc_overrelease_during_dealloc_error` as instructed by the message, and provide the backtrace at that point. It's interesting that the same progress appears to be griping about that a bajillion times in your log - curious if it's once per `upon` block.

Possibly unrelated to this, but repeated use of a weak variable seems to be inviting the variable to poof out from under you in the middle of a block. (That's what `-Wreceiver-is-weak` diagnoses in Obj-C code.) So the block in that sample should probably be doing `guard let update = progress else { return }`. I'd make that fix to your local copy and see if that helps, as well.

zwaldowski:
See if #138 helps.

piercifani:
It looks like I can still make it crash, but it's different now. I managed to create a sample gist:

https://gist.github.com/piercifani/dea9934a143c691a1e422182fe9aaf99

Execute that in the Release mode and you'll see the crash.

zwaldowski:
Remember to remove your observers. I'm not able to reproduce the crash now when `ProgressObserver` looks like:

```swift
class ProgressObserver: NSObject {

    fileprivate let onUpdate: (Progress) -> Void
    fileprivate let progress: Progress

    private static var kvoContext = false

    public init(progress: Progress, onUpdate: @escaping (Progress) -> Void) {
        self.progress = progress
        self.onUpdate = onUpdate
        super.init()
        progress.addObserver(self, forKeyPath: #keyPath(Progress.fractionCompleted), options: .new, context: &ProgressObserver.kvoContext)
    }

    deinit {
        progress.removeObserver(self, forKeyPath: #keyPath(Progress.fractionCompleted), context: &ProgressObserver.kvoContext)
    }

[...]

PR #138, opened 2016-10-28
Making TaskProgress more explicit for multithreaded overrelease (#137)

zwaldowski:

Getting tests to run. Should be a fix for #137.

#### Testing
N/A
#### API Changes
N/A


piercifani:
It looks like I can still reproduce it. I managed to create a sample gist:

https://gist.github.com/piercifani/dea9934a143c691a1e422182fe9aaf99

Execute that in the Release mode and you'll see the crash.

PR #142, opened 2016-10-29
Add a unit test that highlights the error when canceling a task produced using allSucceded

piercifani:

I have found that sometimes a `cancel` message is not propagated correctly thorough the `NSProgress` hierarchy when created with `allSucceeded`. 

#### Testing

A test has been added that highlights this error.

#### API Changes

None


zwaldowski:
Thank you so much! See #145.

PR #146, opened 2016-10-29
Fix deinit races

zwaldowski:
#### What's in this pull request?

Lots of fixes for #144 and friends.

#### Testing

@piercifani has submitted a sample of code as a full-stack test suite. We loop this 20 times to allow as a smoke test for deallocation problems.

#### API Changes

`Task` gains `isCancelled`.

Issue #136, opened 2016-10-22
Review documentation


zwaldowski:
See [current upload from Swift 3 branch](https://bucket.waldowski.me/BNRDeferred-20161022/index.html).

jeremy-w:
Re: [ExistentialTask's definition of](https://github.com/bignerdranch/Deferred/pull/147/files#diff-deebd2f35481182c501a77dd8ea54a50) `init<Task: FutureType where Task.Value: ResultType, Task.Value.Value == SuccessValue>(_ base: Task, cancellation: ((Void) -> Void)? = nil)`:

> It wasn't clear to me from the docs that the cancellation will *only* be called if base is not a Task delivering the same value as Self. That would surprise me for sure as a dev, and I'd probably mess that up at some point. :\ Is it possible to overload such that we don't take the cancellation param unless we're going to use it?

Also includes docs copypasta about base not being a Task, when that's its type in that ctor!

jeremy-w:
NSProgress adopt child ???:

Double negative - so, "i.e., progress is understood to have a parent"?

(I don't understand what this really means, or why we'd mistrust the iOS 9 behavior.)

Return type also is not documented, or why it returns and how it differs from the receiver.

PR #147, opened 2016-10-31
Backport threading fixes from 3.0

zwaldowski:

See #146.

#### Testing

None (including @piercifani's test case from #146).

#### API Changes

None


zwaldowski:
If 👷 is happy, I'll make the necessary changes to release this as 2.2.1. Nerds, please review.

Sources/Task/ExistentialTask.swift
```diff
@@ -91,14 +91,16 @@ extension Task {
         self.init(future: base, progress: progress)
     }
 
+    private typealias _Self = Task<SuccessValue>
+
     /// Create a task whose `upon(_:_:)` method uses the result of `base`.
     ///
     /// If `base` is not a `Task`, `cancellation` will be called asynchronously,
```
jeremy-w: Given `base` has type `Task` in the parameter list, how would we end up with a `base` that isn't a `Task`?

Sources/Task/ExistentialTask.swift
```diff
@@ -91,14 +91,16 @@ extension Task {
...
     /// but not on any specific queue. If you must do work on a specific queue,
     /// schedule work on it.
     public convenience init<Task: FutureType where Task.Value: ResultType, Task.Value.Value == SuccessValue>(_ base: Task, cancellation: ((Void) -> Void)? = nil) {
-        let progress = NSProgress.wrapped(base, cancellation: cancellation)
-        self.init(future: Future(task: base), progress: progress)
+        let underlying = (base as? _Self)?.progress ?? NSProgress.wrapped(base, cancellation: cancellation)
```
jeremy-w: If we find an underlying progress, it looks like we drop the `cancellation` argument on the floor?

Sources/Task/TaskCollections.swift
```diff
@@ -25,13 +25,16 @@ extension CollectionType where Generator.Element: FutureType, Generator.Element.
...
-        let outerProgress = NSProgress(parent: nil, userInfo: nil)
-        outerProgress.totalUnitCount = numericCast(count)
+        let progress = NSProgress(parent: nil, userInfo: nil)
+        progress.totalUnitCount = numericCast(count)
         let group = dispatch_group_create()
 
         for task in self {
-            let innerProgress = NSProgress.wrapped(task, cancellation: nil)
-            outerProgress.adoptChild(innerProgress, orphaned: true, pendingUnitCount: 1)
+            if let task = task as? Task<Generator.Element.Value.Value> {
```
jeremy-w: Is this basically checking that the value types align? How does that tie into orphaned yea or nay? It would be good to make that connection clear in the source.

Sources/Task/ExistentialTask.swift
zwaldowski: Yep.

zwaldowski: It worked that way prior to this commit as well, just less obviously.

Sources/Task/TaskCollections.swift
zwaldowski: The method is overall generic over `Collection where Iterator.Element: FutureProtocol, Iterator.Element.Value: Either, Iterator.Element.Value.Left == Error` (Swift 3, sorry), allowing you to intersperse an array of `Task`s and `Future`s. If it is a `Task<T>`, we want to get its progress/cancellation out. If it isn't, then the sub-task can't be cancelled, but you can still attempt to cancel the overall `Task`.

Sources/Task/ExistentialTask.swift
jeremy-w: It wasn't clear to me from the docs that the cancellation will _only_ be called if `base` is not a Task delivering the same value as `Self`. That would surprise me for sure as a dev, and I'd probably mess that up at some point. :\ Is it possible to overload such that we don't take the `cancellation` param unless we're going to use it?

Sources/Task/TaskProgress.swift
```diff
@@ -29,87 +29,145 @@ private struct KVO {
...
+    init(attachingTo observee: NSProgress) {
+        let current = NSProgress.currentProgress()
+        self.observee = observee
+        super.init(parent: current, userInfo: nil)
+
+        if current?.cancelled == true {
+            observee.cancel()
+        }
+
+        if current?.paused == true {
```
jeremy-w: Do we expect it to be both canceled and paused simultaneously?

Is it legit to `pause` after we've already `cancel`ed?

Sources/Task/TaskProgress.swift
```diff
@@ -29,87 +29,145 @@ private struct KVO {
...
+            observee.pause()
         }
+
+        token = Observation(observing: observee, observer: self)
```
jeremy-w: Perhaps rename `token` to `Observation`?

Sources/Task/TaskProgress.swift
```diff
@@ -120,21 +178,25 @@ extension NSProgress {
     ///
     /// Send `isOrphaned: false` if the iOS 9 behavior cannot be trusted (i.e.,
     /// `progress` is not understood to have no parent).
```
jeremy-w: Double negative - so, "i.e., `progress` is understood to have a parent"?

(I don't understand what this really means, or why we'd mistrust the iOS 9 behavior.)

Return type also is not documented, or why it returns and how it differs from the receiver.

jeremy-w:
LGTM. Looks about as tricky as you'd expect, given the issues at hand.

Sources/Task/ExistentialTask.swift
zwaldowski: It is possible - there's an explicit init-with-`Task<T>`, but it loses out on doing the right thing if you end up with the other initializer. Please drop a comment on #136 mentioning this, if you don't mind.

Sources/Task/TaskProgress.swift
```diff
@@ -29,87 +29,145 @@ private struct KVO {
...
-                resume()
-                pausingHandler = original.pause
+        } else {
+            cancellationHandler = observee.cancel
+        }
+    }
+
+    func inheritPaused(value: Bool) {
+        if value {
+            pausingHandler = nil
```
zwaldowski: Yes, and a couple of other reasons, like cleaning up potential cycles from cancellation and to prevent recursively calling the handler.

Sources/Task/TaskProgress.swift
zwaldowski: Please leave doc comments on #136.

Sources/Task/TaskProgress.swift
```diff
@@ -29,87 +29,145 @@ private struct KVO {
...
 /// tree.
 private final class ProxyProgress: NSProgress {
-    let original: NSProgress
 
-    init(cloning original: NSProgress) {
-        self.original = original
-        super.init(parent: .currentProgress(), userInfo: nil)
-        attach()
-    }
+    /// A side-table object to weakify the progress observer and prevent
```
zwaldowski: Can do.

Sources/Task/TaskProgress.swift
zwaldowski: The behavior we're seeking to replicate from the SDK `-[NSProgress addChild:withPendingUnitCount:]` does exactly this:

```swift
    if ([self isCancelled]) {
        [child cancel];
    }
    
    if ([self isPaused]) {
        [child pause];
    }
```

Just replicating the state machine most faithfully.

Sources/Task/TaskProgress.swift
zwaldowski: I'm fine with the name as it is following the lead of the `NSNotificationCenter` token thingy.

Sources/Task/TaskProgress.swift
zwaldowski: Done. Need to make this change on #146 too.

zwaldowski:
@jeremy-w Remember to 👍 the review when ready, we require approvals for merging to master.

Sources/Task/ExistentialTask.swift
jeremy-w: Comment dropped.

zwaldowski: 😃

Sources/Task/TaskCollections.swift
zwaldowski: Ah, almost forgot this!

jeremy-w:
Oh wait, gotta start a rando review in order to approve. Yeesh.

Sources/Task/TaskCollections.swift
```diff
@@ -29,15 +29,15 @@ extension CollectionType where Generator.Element: FutureType, Generator.Element.
...
+            if let task = future as? Task<Generator.Element.Value.Value> {
                 progress.adoptChild(task.progress, orphaned: false, pendingUnitCount: 1)
             } else {
-                progress.adoptChild(NSProgress.wrapped(task, cancellation: nil), orphaned: true, pendingUnitCount: 1)
+                progress.adoptChild(NSProgress.wrapped(future, cancellation: nil), orphaned: true, pendingUnitCount: 1)
             }
 
             dispatch_group_enter(group)
-            task.upon { result in
+            future.upon { result in
```
jeremy-w: canhaz review

zwaldowski: Nice.

PR #152, opened 2016-11-16
Adds module map to explicitly use Atomics

lyricsboy:

We noticed an issue when using `carthage_cache` to pull in a pre-built Deferred 3.0, where when building our app we'd see compiler errors like this:

`LivingObjectCounter.swift:10:8: Missing required module 'Atomics'`

I learned about module maps and considered that it might be necessary to `use Atomics` in the module map for Deferred. Since that module map was automatically generated, I grabbed the generated version (at least the important bits) and added `use Atomics` to it.

#### Testing

1. Create a project that uses Carthage to build Deferred (pointing at this branch)
1. Build Deferred locally using `carthage bootstrap` or what not.
1. Build the test app and see that it works
1. Use `carthage_cache publish` to put that built product up on S3
1. move away the local Carthage/Build directory
1. `carthage_cache install`
1. build the test app and see that it works

### DIFF
--- a/Configurations/Framework.xcconfig
+++ b/Configurations/Framework.xcconfig
@@ -37,4 +37,4 @@ PRODUCT_NAME = Deferred
 DEVELOPMENT_TEAM = 439592V3LM
 
 // Swift Compiler - Search Paths
-SWIFT_INCLUDE_PATHS = $(inherited) $(SRCROOT)/Sources
+SWIFT_INCLUDE_PATHS = $(inherited) $(SRCROOT)/Sources/**
--- a/Deferred.xcodeproj/project.pbxproj
+++ b/Deferred.xcodeproj/project.pbxproj
@@ -65,6 +65,7 @@
 /* End PBXContainerItemProxy section */
 
 /* Begin PBXFileReference section */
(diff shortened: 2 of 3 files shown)


lyricsboy:
Welp, I thought this was 🏅  but, having submitted it to my Jenkins overlord, it seems more like 🚫 . Still seeing the same error as before.

I tried some more things locally to try to reproduce the build failure: 

* removed built products directory
* built and ran tests using fastlane as is done in the Jenkins environment

Both succeeded where Jenkins is still failing.

zwaldowski:
Hmm, OK. Back to the drawing board. I'm trying a few different ways to reproduce locally. The Atomics module should only exist at compile-time, so I'm not sure why it's being reexported.

If you bang your head against this for a little longer, we can always go back to the approach from 2.x and only use the module for SwiftPM.

lyricsboy:
I tried waving a slightly different rubber chicken, to no avail, inspired by this:

http://stackoverflow.com/questions/31757417/swift-app-missing-required-module-when-importing-framework-that-imports-stati

At this point, we're just going to suffer without carthage_cache, since the Jenkins jobs will succeed if we have to bootstrap carthage there from scratch.

I'm not sure what the real fix is...

zwaldowski:
@lyricsboy Wanna try out `zwaldowski/swift-3_0/atomics-header` on your CI?

lyricsboy:
Yes! Yes I do. Thanks!

PR #153, opened 2016-11-17
Automate project using Fastlane

zwaldowski:

Moves CI configuration and day-to-day out of `.travis.yml` and some scripts all into Fastlane.

Automates some simple versioning tasks that happen from release-to-release.

#### Testing

N/A

#### API Changes

N/A


fastlane/actions/podspec_set_version.rb
```diff
@@ -0,0 +1,64 @@
...
+                                       conflicting_options: [:version],
+                                       conflict_block: proc do |value|
+                                         UI.user_error!("You can only pass either a 'version' or a '#{value.key}', not both")
+                                       end),
+          FastlaneCore::ConfigItem.new(key: :version,
+                                       optional: true,
+                                       description: "Version as version object",
+                                       conflicting_options: [:version],
+                                       conflict_block: proc do |value|
+                                         UI.user_error!("You can only pass either a 'semver' or a '#{value.key}', not both")
```
davidahouse: This should read 'version_number' not 'semver' since that is the param name.

fastlane/actions/podspec_set_version.rb
```diff
@@ -0,0 +1,64 @@
...
+                                       optional: true,
+                                       description: "Version as text",
+                                       conflicting_options: [:version],
+                                       conflict_block: proc do |value|
+                                         UI.user_error!("You can only pass either a 'version' or a '#{value.key}', not both")
+                                       end),
+          FastlaneCore::ConfigItem.new(key: :version,
+                                       optional: true,
+                                       description: "Version as version object",
+                                       conflicting_options: [:version],
```
davidahouse: Think that should be :version_number

fastlane/actions/deferred_readme_set_version.rb
```diff
@@ -0,0 +1,73 @@
+require 'semantic'
+
+module Fastlane
+  module Actions
+
+    class DeferredReadmeSetVersionAction < Action
```
davidahouse: Don't consider this a must fix, but I wonder if this can be made generic enough to work across our other repos? Param could be an array of strings to replace? Not sure if its worth it yet or if we just make it generic once we have another repo that wants to do this.

fastlane/actions/deferred_readme_set_version.rb
```diff
@@ -0,0 +1,73 @@
...
+                                       optional: true,
+                                       description: "Version as text",
+                                       conflicting_options: [:version],
+                                       conflict_block: proc do |value|
+                                         UI.user_error!("You can only pass either a 'version' or a '#{value.key}', not both")
+                                       end),
+          FastlaneCore::ConfigItem.new(key: :version,
+                                       optional: true,
+                                       description: "Version as version object",
+                                       conflicting_options: [:version],
```
davidahouse: should be version_number

fastlane/actions/deferred_readme_set_version.rb
```diff
@@ -0,0 +1,73 @@
...
+                                       conflict_block: proc do |value|
+                                         UI.user_error!("You can only pass either a 'semver' or a '#{value.key}', not both")
```
davidahouse: Also should be version_number

.travis.yml
```diff
@@ -1,65 +1,30 @@
...
+    - script: fastlane ios ci
+      env: JOB=iOS
+    - os: linux
+      dist: trusty
+      sudo: required
+      env: JOB=Linux
+      install:
+        - eval "$(curl -sL https://gist.githubusercontent.com/kylef/5c0475ff02b7c7671d2a/raw/9f442512a46d7a2af7b850d65a7e9bd31edfb09b/swiftenv-install.sh)"
+      script:
+        - swift test
```
davidahouse: Should this belong in the fastfile? Not sure fastlane platforms support linux yet, but this could be a agnostic lane like `linux_ci` or similar.

zwaldowski: It does not yet support Linux as a platform. I too would like it to be in Fastlane, but for now it's what we've got. The script used in the `install` action is intended specifically for use under CI, and isn't suited for a self-run lane.

.travis.yml
```diff
@@ -1,65 +1,30 @@
...
-  else
-    swift test;
-  fi
-after_success:
-- if [[ "$JOB" == "Misc" ]]; then
-    ./Configurations/publish_docs.sh;
-  else
-    sleep 5;
-  fi
+  - fastlane ci
```
zwaldowski: Fair.

fastlane/actions/podspec_set_version.rb
zwaldowski: Yep.

fastlane/Fastfile
```diff
@@ -0,0 +1,61 @@
...
+platform :mac do
+  desc "Test using Swift Package Manager for macOS and Linux"
+  lane :test do
+    sh "swift test"
+  end
+
+  desc "Execute tests, perform CocoaPods linting, publish documentation"
+  lane :ci do
+    test
+    pod_lint
```
zwaldowski: I did some of this, but probably not how you intended. I am adverse to mucking more with Travis matrix stuff (because I don't like them), and I welcome improvements here otherwise.

.travis.yml
davidahouse: `swift test` could potentially be in fastlane, but I'm cool with leaving this as-is. Maybe if we end up with many more commands on linux under the script: tag here then we might reconsider.

zwaldowski: It is not intended that it get more commands. :)

zwaldowski:
@davidahouse Thanks for the feedback!

Issue #149, opened 2016-11-09
Recover and produce a new Task


piercifani:
Right now, there is a way to recover from an error and produce a new default value for the `SuccessValue`. I think a way to recover from an error by producing a new `Task` for the same `SuccessValue` would be helpful.

Do you think something like this would be helpful?

```
extension Task {

   public func recover(upon executor: Executor, start startNextTask: @escaping(Error) -> Task<SuccessValue>) -> Task<SuccessValue> {
       
       let future: Future<TaskResult<SuccessValue>> = andThen(upon: executor) { (result) -> Task<SuccessValue> in
           do {
               let value = try result.extract()
               return Task<SuccessValue>(success: value)
           } catch let error {
               return startNextTask(error)
           }
       }
       
       return Task<SuccessValue>(future: future)
   }
}
```

zwaldowski:
Rest assured I haven't forgotten about this!

I'm not finding much precedent in other functional patterns, and how this would fit in overall. Maybe @jgallagher could weigh in? (I understand I'm asking you a glorified naming question here.)

I'm slightly leaning towards snippetizing this as a pattern in our [wiki](https://github.com/bignerdranch/Deferred/wiki), but if you've found a use for it, it's likely to be valuable as an addition to the project. I just need to think on it a bit more.

piercifani:
Yes, I have found a use for this function, I'm actually shipping it in the next release of an app I'm working on.

Some pseudo-code to help you understand the use-case:

```swift
let fetchOrderTask: Task<PurchaseOrder> = {
    if someBusinessMadeUpCondition() {
        let specialUsersOrderTask = self.fetchOrderForSpecialUsers()
        let recoverFromSpecialUsersOrderTask = specialUsersOrderTask.recover(upon: uponQueue, substituting: { (_) -> Task<PurchaseOrder> in
            return self.fetchNormalOrder()
        })
        
        return recoverFromSpecialUsersOrderTask
    } else {
        return self.fetchNormalOrder()
    }
}()

```

I'm not sure about what functional pattern this represents, I'm not the most knowledged in this regard, but I have found this pretty useful.

Thanks!

zwaldowski:
See #162

PR #159, opened 2016-11-30
Add a new Recover that takes a function that produces a new task

piercifani:
Like mentioned in https://github.com/bignerdranch/Deferred/issues/149, right now, there is a way to recover from an error and produce a new default value for the `SuccessValue`. I think a way to recover from an error by producing a new `Task` for the same `SuccessValue` would be helpful.

I wrote a naive implementation for this requirement but I could not get correctly some features like cancellation and `NSProgress` wrapping.

#### Testing
A unit test for this new method is added to the suite

#### API Changes
A new `recover` method is available

### DIFF
--- a/Sources/Task/TaskRecover.swift
+++ b/Sources/Task/TaskRecover.swift
@@ -47,4 +47,27 @@ extension Task {
         return Task<SuccessValue>(future: future, cancellation: cancel)
         #endif
     }
+
+    public func recoverWithNewTask(upon executor: Executor, start startNextTask: @escaping(Error) -> Task<SuccessValue>) -> Task<SuccessValue> {
+        
+        let deferred = Deferred<TaskResult<SuccessValue>>()
+        upon(executor) { result in
+            switch result {
+            case .success(let value):
+                deferred.fill(with: .success(value))
+            case .failure(let error):
+                let recoverTask = startNextTask(error)
+                recoverTask.upon(executor, execute: { (recoverResult) in
+                    switch recoverResult {
+                    case .success(let value):
+                        deferred.fill(with: .success(value))
+                    case .failure(let error):
+                        deferred.fill(with: .failure(error))
+                    }
+                })
+            }
+        }
+        
+        return Task<SuccessValue>(future: Future(deferred))
+    }
 }
--- a/Sources/Task/TaskRecover.swift
+++ b/Sources/Task/TaskRecover.swift
@@ -50,24 +50,15 @@ extension Task {
 
     public func recoverWithNewTask(upon executor: Executor, start startNextTask: @escaping(Error) -> Task<SuccessValue>) -> Task<SuccessValue> {
         
-        let deferred = Deferred<TaskResult<SuccessValue>>()
-        upon(executor) { result in
-            switch result {
-            case .success(let value):
-                deferred.fill(with: .success(value))
-            case .failure(let error):
-                let recoverTask = startNextTask(error)
-                recoverTask.upon(executor, execute: { (recoverResult) in
-                    switch recoverResult {
-                    case .success(let value):
-                        deferred.fill(with: .success(value))
-                    case .failure(let error):
-                        deferred.fill(with: .failure(error))
-                    }
-                })
+        let future: Future<TaskResult<SuccessValue>> = andThen(upon: executor) { (result) -> Task<SuccessValue> in
+            do {
+                let value = try result.extract()
+                return Task<SuccessValue>(success: value)
+            } catch let error {
+                return startNextTask(error)
             }
         }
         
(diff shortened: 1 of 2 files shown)


zwaldowski:
Seeking naming thoughts and feedback from @jeremy-w. We'll finally get the ball rolling on this. 😺

Sources/Task/TaskRecover.swift
```diff
@@ -47,4 +47,18 @@ extension Task {
...
         #endif
     }
+
+    public func recoverWithNewTask(upon executor: Executor, start startNextTask: @escaping(Error) -> Task<SuccessValue>) -> Task<SuccessValue> {
+        
+        let future: Future<TaskResult<SuccessValue>> = andThen(upon: executor) { (result) -> Task<SuccessValue> in
+            do {
+                let value = try result.extract()
+                return Task<SuccessValue>(success: value)
+            } catch let error {
```
jeremy-w: `} catch {` implicitly binds `error` on our behalf.

Sources/Task/TaskRecover.swift
```diff
@@ -47,4 +47,18 @@ extension Task {
         return Task<SuccessValue>(future: future, cancellation: cancel)
         #endif
     }
+
+    public func recoverWithNewTask(upon executor: Executor, start startNextTask: @escaping(Error) -> Task<SuccessValue>) -> Task<SuccessValue> {
+        
+        let future: Future<TaskResult<SuccessValue>> = andThen(upon: executor) { (result) -> Task<SuccessValue> in
+            do {
+                let value = try result.extract()
```
jeremy-w: As I read it, this puts more weight on the extraction than the wrapping, but I'm not sure that's where we want the focus to go. I'd inline it as `return Task<SuccessValue>(success: try result.extract())`.

Sources/Task/TaskRecover.swift
```diff
@@ -47,4 +47,18 @@ extension Task {
         return Task<SuccessValue>(future: future, cancellation: cancel)
         #endif
     }
+
+    public func recoverWithNewTask(upon executor: Executor, start startNextTask: @escaping(Error) -> Task<SuccessValue>) -> Task<SuccessValue> {
```
jeremy-w: The "recover" naming seems to assume that, unlike the first task, this one is _totally_ gonna pan out. I think that assumption was more warranted in the synchronous case that this is modeled after, but it's possibly misleading here. (And possibly so in the synchronous case too, since that is allowed to `throw` and so fail to recover.)

Some (less optimistic) alternative names:

- failingThat(upon:try:)
- fallback(upon:to:)
- or(upon:execute:)

These also improve readability in the event you have several fallbacks available. I dunno about you, but seeing a chain of recover with this, recover with that, no wait, recover with this other thing, would leave me a bit confused - if it legit recovers, you should only ever need one, right?

jeremy-w: Totally not in this PR, but since I saw it while looking at this PR - there's a typo in the recover(upon:substituting:) variant here:

> The resulting task is cancellable in the same way the **recieving** task is.

jeremy-w: (A spellcheck lint step for comments would be pretty neat.)

Tests/TaskTests/TaskTests.swift
```diff
@@ -266,5 +266,21 @@ class TaskTests: CustomExecutorTestCase {
...
     }
+    
+    func testThatRecoverAlsoProducesANewTask() {
+        let expectation = self.expectation(description: "recover produces a new task")
+        let task: Task<Int> = anyFailedTask.recoverWithNewTask(upon: executor) { _ in
+            return self.anyFinishedTask
+        }
+        
+        task.upon {
+            XCTAssertEqual($0.value, 42)
```
jeremy-w: The "42" kind of comes out of nowhere and leaves me confused when reading the test. Maybe assert `$0.value` == `anyFinishedTask.value` to make the connection clear?

Sources/Task/TaskRecover.swift
piercifani: Agreed that the naming is not ideal here. 

I really like `fallback(upon:to:)` so if I don't see any other suggestions I'd change it for this option.

Sources/Task/TaskRecover.swift
piercifani: Yep, I always tend to be more verbose than I should, changing that right away!

Tests/TaskTests/TaskTests.swift
piercifani: Agreed

zwaldowski: Not his fault, the `recover` test is faulty in the same way. I'll make a pass at that soon.

Issue #166, opened 2017-01-12
Swift Package Manager support


mdmathias:
If following [the docs](https://bignerdranch.github.io/Deferred/getting-started.html), you will end up with a `Package.swift` like so:

```swift
import PackageDescription

let package = Package(
    name: "My Extremely Nerdy App",
    dependencies: [
        .Package(url: "https://github.com/bignerdranch/Deferred.git", majorVersion: 3),
    ]
)
```

Then, using `swift package fetch` in Terminal.app, you will see this error:

> error: The dependency graph could not be satisfied. The package (git@github.com:bignerdranch/Deferred.git) with version tag in range (3.0.0..<3.9223372036854775807.9223372036854775807) is not found. Found tags ([2.0.0-rc.1, 2.0.0, 2.2.0-rc.1, 2.2.0, 2.2.1, 3.0.0-beta.1, 3.0.0-beta.2, 3.0.0-beta.3, 3.0.0-rc.1, 3.0.0-rc.2])

Okay...so I will explicitly specify a `Version` that (I think) has a resolvable release tag.

```swift
import PackageDescription

let package = Package(
    name: "My Extremely Nerdy App",
    dependencies: [
        .Package(url: "https://github.com/bignerdranch/Deferred.git", Version(2,2,1)),
    ]
)
```

But then I get:

[...]

PR #167, opened 2017-01-12
Fix README replacement for prerelease versions

zwaldowski:
Addresses #166. NFC.


zwaldowski:
👷 is blocked. Mergin'

PR #174, opened 2017-03-28
Fix folder layout for Swift Package Manager

zwaldowski:
SwiftPM 3.1 changed the package format under the assumption that the previous ones were wrong.

Whoops.

Given that it's likely my error, the intent here is to re-tag 3.0-rc back to 3.0.0 with this fix.


zwaldowski:
Waiting to see what 👷 says about this on 🐧.

PR #176, opened 2017-04-07
Move TaskResult to Task.Result

masgharneya:
This change moves TaskResult into Task.Result. This is only possible on Swift 3.1 and up so the old TaskResult exists where it needs to. Tested out using 3.0.2 and 3.1 toolchain. 

Wasn't sure we we wanted to mark
```swift
public typealias TaskResult<Value> = Task<Value>.Result
```
as deprecated yet. Should we go ahead and do that?


masgharneya:
Added the deprecation attribute. I was able to resolve all but one of the warnings. Not sure how to best resolve the last one (`extension TaskResult: Either`).
The only options I see are either living with the warning until we remove `TaskResult` or 
```
#if swift(>=3.1)
extension Task.Result: Either { ... }
#else
extension TaskResult: Either { ... }
#endif
```
but then we are going to end up just copy and pasting the same code for both.

zwaldowski:
Hm. That's annoying. Yeah, I lean slightly towards duping the code since it's only a few lines.

Could we define `private typealias _TaskResult`? Can we extend that publicly?

masgharneya:
> Non-nominal type '_TaskResult' cannot be extended
I get that compiler error on swift 3.0

zwaldowski:
Ugh, yeah, you're right. That duplication is ugly. Oh well, it's fine.

Sources/Task/TaskResult.swift
```diff
@@ -7,14 +7,23 @@
...
+    case success(SuccessValue)
     /// Contains the error value
     case failure(Error)
 }
+#endif
+
+#if swift(>=3.1)
+extension Task.Result: Either {
+
+    public typealias Value = SuccessValue
```
zwaldowski: Add an `@available(*, deprecated, renamed: "SuccessValue")` to both of these. Might get some more warnings from that.

Sources/Task/TaskResult.swift
```diff
@@ -7,14 +7,23 @@
 //
 
 /// An enum for returning and propogating recoverable errors.
```
zwaldowski: Move this doc comment into the 3.0 block, neither types have documentation with this placement.

zwaldowski:
Excellent! Thank you!

zwaldowski:
I do want @mdmathias to give it a once-over for his book but I'm 👍.

mdmathias:
Thanks for the ping. I will get to this on Monday. I'm sure the code is 👌 Merge if y'all are happy and I'll make sure to update our book/course.

PR #183, opened 2017-05-15
Escape documentation filenames for documentation

zwaldowski:
Discovered by @wookiee. See realm/jazzy#547.


zwaldowski:
Tested locally

PR #190, opened 2017-07-19
Crash fix where nil block can be called

zwaldowski:

Fixes an issue where mutation of callback handlers in-situ can cause an attempt at calling an Objective-C block that is `NULL` by code in Foundation, which is fatal.

#### Testing

The issue has been resilient to non-contrived reproduction.

#### API Changes

Should be no compatibility changes.


piercifani:
Any ETA on this?

zwaldowski:
I’ll merge and tag a release this morning.

Issue #160, opened 2016-12-15
Integrate playground into repository


zwaldowski:
- [ ] `.playground` that ships along `Deferred.xcworkspace`.
- [ ] Fastlane script that takes that `.playground` and makes it into an independent `.playground`
 - [ ] Fastlane action that takes that `.playground` and makes it into an independent `.playgroundbook`

zwaldowski:
See #161

zwaldowski:
A next step will do this will be to canonicalize the lock-free shims in the absence of the Xcode project/Atomics pseudomodule. Apple's libc ships with the methods we'd need ([1](https://github.com/apple/swift-compiler-rt/blob/swift-4.0-branch/lib/builtins/atomic.c), [2](https://github.com/apple/swift-compiler-rt/blob/swift-4.0-branch/lib/builtins/atomic_flag_clear.c), [3](https://github.com/apple/swift-compiler-rt/blob/swift-4.0-branch/lib/builtins/atomic_flag_test_and_set.c)), which I've had success with integrating already.

zwaldowski:
Closing due to inactivity.

PR #215, opened 2018-04-02
Upstream playground atomics shims

zwaldowski:

Codifies the hacks done for the [Playground](https://github.com/bignerdranch/Deferred/wiki/Playground) so that they can be tested under CI.

#### Testing

CI will build but will not (for now) execute the framework in a pseudo-Playgrounds-compatible mode.

#### API Changes

N/A. Internal bits only.


zwaldowski:
Whoops, also note this works towards #160.

PR #217, opened 2018-04-02
Return an array from Task allSucceeded() as expected.

zwaldowski:

Implements a request made so that  `allSucceeded()` for Task aligns with `allFilled()` for Future.

#### Testing

The method still executes largely the same code.

##### TODO

- [ ] Check the return value.

#### API Changes

This is an API change that may be source breaking to clients, as the return type of the task changes. You may call `ignore()` on it instead. If you're just calling `upon`, Swift 4 might've forced you to already insert `_ in`, so this would not affect that.


zwaldowski:
I think the way forward is to come up with a new name and have `allSucceeded()` return `newThing().ignored()`.

PR #222, opened 2018-04-28
Update for Swift 4.1 on Linux

zwaldowski:

Cherry-picks from #217.

#### Testing

CI should cover us here.

#### API Changes

Nothing public-facing, for now.


zwaldowski:
Should probably also run CI against old versions, as long as we're supporting Swift 4.x.

Issue #203, opened 2018-02-27
Allow printing the handlers attached to Deferred while debugging


zwaldowski:
I'd like to see this through `debugDescription` or `Mirror` (i.e., `po` output in the debugger). This would require a representational change to move away from `DispatchGroup` for storing handlers, which I already have a successful prototype of. But even then it's not full-fidelity; handlers `print` as "`(Function)`". LLDB is best equipped to get the actual name and source of a closure, so a custom debugger command might work?

zwaldowski:
See "[LLDB Data Formats](https://lldb.llvm.org/varformats.html)."

zwaldowski:
See "[Webkit/lldb_webkit.py](https://github.com/WebKit/webkit/blob/master/Tools/lldb/lldb_webkit.py)".

zwaldowski:
WWDC18's announcements serve possibly serve as a more complete solution: 

https://twitter.com/shaps/status/1003721647315148801

zwaldowski:
This is a pipe dream for now. The information here is good as an archive and I'll reopen it if I have any ideas.

PR #229, opened 2018-04-29
Skinny Deferred

zwaldowski:
`Deferred` is meant to handle large parts of an app's asynchronous control flow, so its performance is critical. We've strived in the past to use the best constructs we have at our disposal to make this possible, like being early adopters of GCD in Swift. #213 revealed we can still do better! 😃

### What's in this pull request?

By combining our waiters queue and the storage, `Deferred` (the type) can reach same-ballpark performance as GCD while maintaining the same invariants.

Per the performance tests, this PR features:

- A 70% reduction in peak memory usage
- A 35% speed improvement in simple operations
- An 80% speed improvement in chained operations
- Smaller call stacks when debugging `upon` handlers.

In summary: same great taste, less filling.

This representation allows Swift-side access to handlers, possibly allowing for #203.

#### TODO

- [x] What about code size? This PR gains about 30KB for Release under 4.1, which is livable, but not ideal for my OCD.

### Testing

Largely no impact, but `Deferred` gains a specialization `where Value: AnyObject` that needs testing.

#### TODO

- [x] Needs duplicated tests to handle `AnyObject` specialization
- [x] Double-check other coverage implications

### API Changes

No public API changes here, it's all under the hood.

### DIFF
--- a/Sources/Deferred/Deferred.swift
+++ b/Sources/Deferred/Deferred.swift
@@ -8,8 +8,17 @@
 
 import Dispatch
 
-/// A deferred is a value that may become determined (or "filled") at some point
-/// in the future. Once a deferred value is determined, it cannot change.
+/// A value that may become determined (or "filled") at some point in the
+/// future. Once determined, it cannot change.
+///
+/// You may subscribe to be notified once the value becomes determined.
+///
+/// Handlers and their captures are strongly referenced until:
+/// - they are executed when the value is determined
+/// - the last copy to this type escapes without the value becoming determined
+///
+/// If the value never becomes determined, a handler submitted to it will never
+/// be executed.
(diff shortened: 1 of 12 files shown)


Sources/Deferred/Deferred.swift
```diff
@@ -8,133 +8,77 @@
...
+        } ?? handler(value)
     }
+}
 
-    static func convertFromReference(_ value: AnyObject) -> Value {
-        return unsafeDowncast(value, to: Box.self).wrapped
+extension Deferred: PromiseProtocol {
+    @discardableResult
+    public func fill(with value: Value) -> Bool {
+        return variant.awake(with: value)
```
zwaldowski: Rename `store(_:)`.

Issue #224, opened 2018-04-28
Adopt conditional conformances


zwaldowski:
Probably just for `Task.Result`, but maybe it's time for `TaskProtocol` to see what else we can improve.

PR #253, opened 2018-09-17
Move Task functionality into conditional conformances

zwaldowski:

Working towards #224.

Extracts `extension Task` and `extension FutureProtocol where BlahBlahBlah` into `extension TaskProtocol`. It resolves the major ergonomic issues with `Task` methods, allowing you to call the `map`, `andThen`, etc. you're expecting to on `Future<TaskResult<T>>` without wrapping it as `Task`. Beyond that, the idea is to reduce complexity in the API and to reduce the occurrence of compiler errors from not extending the right thing. 

#### Testing

None intended.

#### API Changes

Many, but they _should_ be source-compatible.


Sources/Task/ExistentialTask.swift
```diff
@@ -57,15 +62,15 @@ public final class Task<SuccessValue>: NSObject {
     }
 
     /// Creates a task whose `upon(_:execute:)` methods use those of `base`.
-    public convenience init<Task: FutureProtocol>(_ base: Task, progress: Progress)
-        where Task.Value: Either, Task.Value.Left == Error, Task.Value.Right == SuccessValue {
-        self.init(future: Future(task: base), progress: progress)
+    public convenience init<OtherTask: TaskProtocol>(_ base: OtherTask, progress: Progress) where OtherTask.SuccessValue == SuccessValue {
```
zwaldowski: These go back and forth a few times in the project; not sure if the generic parameter should be `OtherTask` or just `Other`.

Sources/Task/TaskAsync.swift
```diff
@@ -6,10 +6,10 @@
 //  Copyright © 2015-2018 Big Nerd Ranch. Licensed under MIT.
 //
 
+import Dispatch
```
zwaldowski: Rewrite this out.

Sources/Task/TaskProgress.swift
```diff
@@ -201,13 +201,12 @@ extension Progress {
         let progress = Progress(totalUnitCount: 0)
         progress.completedUnitCount = 1
         progress.isCancellable = false
-        progress.isPausable = false
```
zwaldowski: Remove this, not related to this PR

Sources/Task/TaskProgress.swift
```diff
@@ -201,13 +201,12 @@ extension Progress {
...
         return progress
     }
 
     /// A simple indeterminate progress with a cancellation function.
     static func wrappingCompletion<OtherFuture: FutureProtocol>(of base: OtherFuture, cancellation: (() -> Void)?) -> Progress {
-        let totalUnitCount: Int64 = base.wait(until: .now()) != nil ? 0 : -1
```
zwaldowski: Split this out, not related to this PR

Sources/Task/TaskPromise.swift
```diff
@@ -0,0 +1,47 @@
...
+}
+
+extension TaskProtocol where Self: PromiseProtocol, SuccessValue == Void {
+    /// Completes the task with a success.
+    ///
+    /// Fulfilling this deferred value should usually be attempted only once.
+    ///
+    /// - seealso: `PromiseProtocol.fill(with:)`
+    /// - seealso: `TaskProtocol.succeed(with:)`
+    @discardableResult @available(swift 4)
```
zwaldowski: We can probably drop availability attributes.

zwaldowski: Will do later…

Issue #260, opened 2018-10-10
How to correctly track a Task's progress?


piercifani:
Hi there!

I've written a URLSession extension that allows you to upload a file and wrap it in a task like this:

```
extension URLSession {

  public func uploadFile(with urlRequest: URLRequest, fileURL: URL) -> Task<APIClient.Response> {
    
    let deferred = Deferred<Task<APIClient.Response>.Result>()
    let uploadTask = self.uploadTask(with: urlRequest, fromFile: fileURL) { (data, response, error) in
      self.analyzeResponse(deferred: deferred, data: data, response: response, error: error)
    }
    uploadTask.resume()
    if #available(iOS 11.0, tvOS 11.0, watchOS 4.0, macOS 10.13, *) {
      return Task(deferred, progress: uploadTask.progress)
    } else {
      return Task(deferred, cancellation: { [weak uploadTask] in
        uploadTask?.cancel()
      })
    }
  }
}
```

As you see, I'm using the `URLSessionUploadTask`'s progress to initialise the Task. After that, I'm using `andThen` to parse the JSON response from the server and more cleanup.

However, when observing the progress, the upload task is merely a fraction of the total progress, not corresponding to the much more heavier task that it represents.

How should we proceed in such cases? I realise this is more a `NSProgress` question than a Deferred question, but maybe someone can shed some light on this. You can see this behaviour in [this commit](https://github.com/theleftbit/BSWFoundation/commit/c2b86c9d3b7ad7ea4002bac24bac7da4d657d19f) of my project which is based on BNRDeferred

BTW: Really looking forward to version 4.0 of Deferred 💪

piercifani:
Whoops, just saw https://github.com/bignerdranch/Deferred/pull/259, I'll be following that instead

piercifani:
BTW: I updated my project to use the latest from master and I am seeing a difference in progress report, however the change is not exactly what I'm looking for: The `fractionCompleted` property is starting at 0.8 and then climbing all the way to 1.0, but it's still not yet correct IMHO. Maybe I'm doing something wrong?

zwaldowski:
The snippet you've provided reflects something I'd expect #259 to track correctly, as it was made while debugging an identical upload task (ping @DrewFitz?).

zwaldowski:
Thanks, that's quite helpful!

[...]

PR #264, opened 2018-11-10
Bump deployment targets to iOS 9-era releases

zwaldowski:
Reworking `Task` progress to address #260 will require correctly (finally!) using explicit progress composition through [`Progress. addChild(_:withPendingUnitCount:)`](https://developer.apple.com/documentation/foundation/progress/1417260-addchild). We probably could've gone ahead and done that 2 years ago, so it's definitely time. Removes the associated low-hanging fruit of backwards compatibility code, pending larger changes soon.


zwaldowski:
OK, perhaps not…

Issue #271, opened 2019-01-08
Align Task.Result with Swift 5 Result


zwaldowski:
This should happen before Swift 5.0 releases so users of Deferred 4.0 can have the easiest transitions possible.

PR #274, opened 2019-01-13
Align Task and Task.Result with Swift 5 Result

zwaldowski:

Completes #271.

* `Either.Left` is no longer an associated type. Post-Swift 5, we can add the associated type but not change one, so we leave it as an alias for `Error`.
* `Task.Result.extract()` becomes `get()`.
* `Task.Result.withValues(ifLeft:ifRight:)` is removed in favor of `get()`.
* `Task.Result.init(from:)` becomes `Task.Result.init(catching:)`.
* `Task.Result.map(_:)` and `Task.Result.flatMap(_:)` no longer do any magic around `throws`.
* Adds `Task.Result.mapError(_:)` and `Task.Result.flatMapError(_:)`.
* `Task<SuccessValue>` becomes `Task<Success>`, including `Task.Result`.

#### Testing

Improves coverage of `TaskResult` and combines together its scattered sources of testing.

#### API Changes

Yes, it does. Details above. Here, we're supporting upgrading from Deferred 3 with `@availability`, but not previous betas (sorry).

The intent here is that, although we plan to release Deferred 4 soon, we want to ease pain for upgrading to both it and the upcoming Swift 5.


Sources/Task/Either.swift
```diff
@@ -3,56 +3,59 @@
...
+    /// a success, or to catch the value if it is left-biased.
+    func get() throws -> Right
 }
 
-extension Either where Left == Error {
-    /// Derive a success value by calling a failable function `body`.
+extension Either {
     @_inlineable
-    public init(from body: () throws -> Right) {
+    public init(catching body: () throws -> Right) {
```
zwaldowski: Comment now lives in the protocol body.

PR #284, opened 2019-04-08
Fixes issue where multiple nested andThen did not track progress (#281)

### LINKED ISSUE #281: Using a Task's progress after an `andThen`
Hey guys,

I'm again having some issues when observing a Task's progress if the larger chunk of it it's done in an `andThen` block. In the following snippet, the `fractionCompleted` of the outer Task (called `finalTask`) jumps from 0.057 to 0.96 and stalls there until the bigger chunk is completed. However the `uploadTask`'s progress is correctly tracked.

```swift
    func testGetAndUpload() throws {

        let getTask = sut.fetchData()
        let finalTask = getTask.andThen(upon: .main) { _ -> Task<VoidResponse> in
            let uploadTask = self.sut.uploadVideo(videoURL)
            var uploadProgress: ProgressObserver! = ProgressObserver(progress: uploadTask.progress) { (progress) in
                print("UploadTask \(progress.fractionCompleted)")
            }
            uploadTask.upon(.main, execute: { (_) in
                uploadProgress = nil
            })
            return uploadTask
        }

        var completeTaskProgress: ProgressObserver! = ProgressObserver(progress: finalTask.progress) { (progress) in
            print("CompleteTask \(progress.fractionCompleted)")
        }
        let _ = try self.waitAndExtractValue(finalTask, timeout: 10)
        completeTaskProgress = nil
    }
```

The previous code is extracted from the `testGetAndUpload` in 
https://github.com/theleftbit/BSWFoundation/tree/progress-reporting, feel free to clone and execute this test to check on the command line the results.

zwaldowski:
Closes #281.


piercifani:
I checked It out on my project and it looks much much better.

Thanks!

piercifani:
I don't know if this is ready yet, but checking it out on my local project doesn't seem to show any improvements

Here is the log output of the main `Task`:
```
Test Case '-[Tests.APIClientTests testGetAndUpload]' started.
CompleteTask 0.009523809523809525
CompleteTask 0.05714285714285714
CompleteTask 0.06619047619047619
CompleteTask 0.9619047619047619
``

piercifani:
True, it was Cocoapods not linking the right version... sorry about that!

The only weird thing now is that it looks like `fractionCompleted` goes all the way to 0.99 and then moves down to 0.5, that might be strange for the user...

PR #286, opened 2019-05-03
Transition to Circle CI

zwaldowski:

New CI setup. We use Circle internally because it is cost effective and much, much, _much_ faster.

#### Testing

No code changes. The overall shape of the testing matrix doesn't change. Some actions are moved around in an attempt to reduce time spent.

#### API Changes

N/A


.circleci/config.yml
```diff
@@ -0,0 +1,130 @@
...
+  ruby-bundle:
+    commands:
+      install:
+        description: "Install Ruby dependencies using cache"
+        steps:
+          - restore_cache:
+              key: &cache_key 6-gems-{{ checksum "Gemfile.lock" }}-{{ .Environment.CIRCLE_JOB }}
+          - run:
+              name: Install Ruby dependencies
+              command: bundle check --path=~/gems || bundle install --deployment --jobs=3 --path=~/gems --retry=3
```
zwaldowski: Need to annotate

PR #306, opened 2020-10-07
Apple atomic

eqwalker:
Change Deferred Swift package to use swift-atomic dependency:
https://github.com/apple/swift-atomics.git

This change is part of a larger effort to build the BNR iOS template in XCode 12.
after encountering the dreaded 'compiled with X.X version of Swift Compiler' message with the carthage frameworks, I decided to try Swift Package Manager... Luckily Defer was already defined as a Swift Package so I updated the version of  swift atomic and got it testing & building.


eqwalker:
want to squash commits

PR #307, opened 2020-10-07
Replace Deferred's CAtomics with Apple's swift-atomics package

eqwalker:
Change Deferred Swift package to use swift-atomic dependency:  https://github.com/apple/swift-atomics.git

This change makes Deferred compatible with Xcode 12 when used as a Swift package.

After encountering the dreaded 'compiled with X.X version of Swift Compiler' message with the carthage frameworks, I decided to try Swift Package Manager. Deferred was already defined as a Swift Package, so I updated the version of swift atomic and got it testing & building.

#### Testing
No changes. Tests run

#### API Changes
No changes

### DIFF
--- a/.swiftpm/xcode/package.xcworkspace/contents.xcworkspacedata
+++ b/.swiftpm/xcode/package.xcworkspace/contents.xcworkspacedata
@@ -1,0 +1,7 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<Workspace
+   version = "1.0">
+   <FileRef
+      location = "self:">
+   </FileRef>
+</Workspace>
--- a/Package.resolved
+++ b/Package.resolved
@@ -1,0 +1,16 @@
+{
+  "object": {
+    "pins": [
+      {
+        "package": "swift-atomics",
+        "repositoryURL": "https://github.com/apple/swift-atomics.git",
+        "state": {
+          "branch": null,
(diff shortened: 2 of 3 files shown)


zwaldowski:
Apple's package is not a drop-in replacement for the current bridging; the uses in the code will have to be changed as well. The Mac tests are failing because of this. The only reason the iOS tests aren't failing is because the iOS tests test the Xcode project, so it's not seeing any of this change anyway.

The framework should be able to used via SPM already, no change needed. If it's not, please share what integration issues you're having.

Apple's new package is not considered stable at this time, so I'm not sure we can safely drop it in anyway, never mind getting it to integrate with Carthage and CocoaPods. (Keep in mind, that to support all our clients, this project supports all three integration mechanisms.)

eqwalker:
@zwaldowski It's interesting because if you open the Package.swift file and use the XCode generated workspace for packages it builds & tests for MacOS, iOS, tvOS schemes. Not for Apple Watch tho. Maybe we have to change CircleCI to build using Package.swift? I don't yet have the knowledge of CircleCI to say. I do see the build & tests failing when I use the current project as you describe. Long term would Carthage & CocoaPods be an issue if SPM works?

zwaldowski:
> Long term would Carthage & CocoaPods be an issue if SPM works?

Yes, it would.

eqwalker:
Not needed

